Issue #10, opened 2019-03-15
Mobile menu


jankeromnes:
Part of #2.

I'd like to answer these questions:

1. Do we want an always-visible call-to-action button in the mobile menu bar? (E.g. "Login")
2. How to layout our mobile menu?
3. Do we want a second menu for Docs, or a single merged menu?

Here are a few mobile sites that are comparable to Gitpod, for inspiration.

---

<img width="640" alt="Screenshot 2019-03-15 at 19 03 57" src="https://user-images.githubusercontent.com/599268/54452329-36482200-4755-11e9-97a4-eab78c9ff802.png">

- Right menu icon
- Partial overlay menu
- Right call to action button: "Sign up"

---

<img width="640" alt="Screenshot 2019-03-15 at 18 54 53" src="https://user-images.githubusercontent.com/599268/54452419-68f21a80-4755-11e9-959b-2d1636da78f6.png">

- Left menu icon
- Full overlay menu
- Right call to action button: "Go to app"

---

<p align="center">
<img width="320" alt="Screenshot 2019-03-15 at 18 56 57" src="https://user-images.githubusercontent.com/599268/54453089-db172f00-4756-11e9-8b99-134373acc32b.png">
</p>

<img width="640" alt="Screenshot 2019-03-15 at 18 57 16" src="https://user-images.githubusercontent.com/599268/54453105-e4080080-4756-11e9-959b-e2baed870d65.png">

- Two menu icons: Left for Docs tree, Right for website
- Full overlay menus
- (No call to action button, but Kubernetes is not an app)

jankeromnes:
FYI we already have a few concepts:

<img width="320" alt="Button use – 1" src="https://user-images.githubusercontent.com/599268/54453574-e4ed6200-4757-11e9-9315-98e9788966ba.jpg">

- Right menu icon -- blue when open (I'd prefer the icon to become a cross)
- Full overlay menu -- non-standard grid of boxes (I'd prefer a familiar list)
- No call to action button (I'd prefer an always-visible "Login" button)

---

<img width="320" alt="Docs" src="https://user-images.githubusercontent.com/599268/54453596-f0d92400-4757-11e9-8839-59e49f09d34d.jpg">

- Two menu icons on the right... confusing, but tricky to get right

Can we make the docs menu less confusing? E.g. move the Docs menu icon to the left, or merge the docs tree into the common menu? Not sure.

ChristinFrohne:
[...]

PR #11, opened 2019-03-18
[WIP][mobile] implemented mobile menu

jankeromnes:
Implements #10.

Currently looks like this (iPhone 6/7/8 width):

Screenshot 2019-03-18 at 14 31 15

Still some work left to do (e.g. fix gaps around menu, replace menu icon with cross when open) but I'd love early feedback.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -3,9 +3,11 @@ import styled from '@emotion/styled'
 import { transparentize } from 'polished'
 import { Link } from 'gatsby'
 
-import { heights, dimensions, colors } from '../styles/variables'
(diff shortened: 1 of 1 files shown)


jankeromnes:
Thanks a lot for your helpful suggestions! I've implemented those related to the menu (I'll fix each page individually afterwards, including the index page, so I'll keep your suggestions for that in mind).

It looks like this now:

Screenshot 2019-03-18 at 17 14 53

Notes:
- No search icon, as we don't have a search feature yet
- The blue hover effect on Docs doesn't exist on mobile (but links may appear blue when you tap them, which I guess is ok)

I believe it's good enough for merging, but we can definitely iterate on it some more.

PR #14, opened 2019-03-19
[mobile] fix docs page

jankeromnes:
Expandable table of contents in mobile menu (I also include the ToC at the bottom of the mobile Docs page, for convenience):

Screenshot 2019-03-19 at 16 53 04

---

Docs page on mobile (needs a bit more padding at the top, but otherwise quite readable):

Screenshot 2019-03-19 at 16 54 10

---

Docs page on mobile in landscape mode (a bit more padding left and right -- iPad in portrait looks the same):

Screenshot 2019-03-19 at 16 54 44

---

iPad in landscape mode is able to show the full page (non-mobile version).

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -20,8 +20,7 @@ const StyledHeader = styled.header`
     z-index: 2;
     position: fixed;
     width: 100%;
-    padding: 0;
-    padding-top: 5px;
+    padding: 5px 0;
     margin: 0;
   }
 `
@@ -38,9 +37,13 @@ const HeaderInner = styled(Container)`
 
 const MobileMenu = styled.div`
   margin-left: auto;
+  margin-right: -10px;
+  display: flex;
+  flex-direction: row;
+  align-items: start;
 
   > :not(:last-child) {
-    padding-right: 15px;
+    padding-right: 12px;
   }
 
   @media (min-width: ${getEmSize(breakpoints.md)}em) {
@@ -64,6 +67,8 @@ const Menu = styled.div`
     width: 100%;
     margin-top: 15px;
     flex-direction: column;
+    max-height: 85vh;
+    overflow: scroll;
 
     :before {
       content: '';
@@ -99,6 +104,15 @@ const Menu = styled.div`
   }
 `
 
+const HomepageLogo = styled.img`
+  width: 120px;
+
+  @media (max-width: ${getEmSize(breakpoints.md)}em) {
+    width: 100px;
+    margin-top: 3px;
+  }
+`
+
 const HomepageLink = styled(Link)`
   color: ${colors.fontColor1};
   font-size: ${dimensions.fontSize.menu};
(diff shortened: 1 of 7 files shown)


jankeromnes:
The up/down arrow is actually the left arrow rotated. It has a strange blue tint.

ChristinFrohne:
looks good! 
- the spacing in the menu is better now. But i would still give it more space above the logo. 
- any reason why you left out the other sites like 'contact' or 'about' in the menu? Is it too much for now and we will add them another time? 
- What is the problem with the arrow? Can you not get rid of the blue glow?
- I would use a smaller 'log in' button on the mobile screen
- for the doc screen: give it some more space to the sides and the top

jankeromnes:
> looks good!

Thanks!

> the spacing in the menu is better now. But i would still give it more space above the logo.

Ok, I'll try to make the top bar taller to give it more space.

> any reason why you left out the other sites like 'contact' or 'about' in the menu?

I just used the links that were already visible in the non-mobile menu.

- However, I can add an About link (should it also be visible in the non-mobile menu?)
- There is no Contact page. Should I add one?
- Note: The footer has links to Features, Pricing, Docs, Blog, About, Careers, Status, Enterprise. (Should I add a link to Contact there as well?)

> What is the problem with the arrow? Can you not get rid of the blue glow?

Yes the problem is that it's a bit blue. I should probably remove the blue glow with CSS.

> for the doc screen: give it some more space to the sides and the top

I totally agree with more space at the top, however I feel like increasing the padding on the sides is not ideal, because mobile screens are already so narrow (too much padding will make each page very very long).

I'll try to read through the docs on an actual phone to see if we need more left/right padding.

svenefftinge:
I first didn't find the menu. Only after reading the comment here I figured it was nested within the main menu. Not a good sign.

The resizing isn't good for other sizes. E.g.:

Screenshot 2019-03-20 at 09 01 30

jankeromnes:
> I first didn't find the menu.

Thanks for the feedback. There are two docs menu on mobile: expandable in mobile menu, and at the bottom of each docs page.

Please suggest a better place (or places).

jankeromnes:
> The resizing isn't good for other sizes. E.g.:
> 674 x 710

Is that a common window size? I usually test for mobile (375 x 667), mobile landscape (667 x 375), iPad (768 x 1024) and iPad landscape (1024 x 768).

jankeromnes:
> Actually it doesn't look right in any size. Did you try using the responsive mode in the dev tools?

Yes, I'm using Firefox Devtools' reponsive design, in which the website looks great (and also on a real device).

I've also tried Chrome Devtools responsive tool, and was able to reproduce your bug. It's because of the Gitpod logo all the way to the right. I'll remove it.

jankeromnes:
- [x] rebased
- [x] adjusted mobile menu bar sizing
- [x] added About and Contact links (only on mobile, also currently gives "404 page not found")
- [x] fixed Chrome-only overflow bugs
- [x] increased top padding of Docs content on mobile

jankeromnes:
New look in Firefox:

Screenshot 2019-03-20 at 10 35 30

And in Chrome:

Screenshot 2019-03-20 at 10 38 33

svenefftinge:
Cool, looks much better! 👍 

In this size there is no menu:
Screenshot 2019-03-20 at 11 40 15

svenefftinge:
We should disable the background logos.
Screenshot 2019-03-20 at 11 42 43

svenefftinge:
For the menu I like how netlify do it in a separate combo box.

Screenshot 2019-03-20 at 11 44 43
Screenshot 2019-03-20 at 11 44 49

jankeromnes:
Shouldn't `yarn` automatically run `npm install`?

While trying to open this PR in Gitpod, it ran `yarn` in pre-build, but `yarn dev --host 0.0.0.0` fails with:

> error There was a problem loading the local develop command. Gatsby may not be installed. Perhaps you need to run "npm install"?
> Error: Cannot find module 'toidentifier'

However, if I manually run `npm install` (takes a long time) then afterward everything works.

Potentially related, a bunch of files are considered "Deleted" while I haven't done anything in that workspace yet:

Screenshot 2019-03-20 at 14 48 10

jankeromnes:
- [x] rebased
- [x] hide all background logos on mobile
- [x] fixed off-by-one error in responsive breakpoints
- [x] added docs topic selector

Firefox:

Screenshot 2019-03-20 at 18 52 11

Chrome:

Screenshot 2019-03-20 at 18 53 14

svenefftinge:
Cool! Is it on purpose that the combo box only contains the first level of headings?

jankeromnes:
Yes (top pages then have sub-links) but I'm happy to add all entries if you think that's better.

svenefftinge:
With a little indentation I think it would be simpler. Also there is still something in the main menu.

jankeromnes:
> With a little indentation I think it would be simpler.

Sure, will try.

> Also there is still something in the main menu.

You want the expandable docs tree removed from the main menu? (I thought it was somewhat useful yet discreet enough to not be a problem.)

jankeromnes:
Screenshot 2019-03-21 at 09 27 16

src/components/Header.tsx
```diff
@@ -115,38 +130,69 @@ interface HeaderProps {
...
-                    <button onClick={this.toggleMenu} style={{border: 0}}>
+                    <button onClick={this.toggleMenu} style={{ border: 0, padding: '6px 10px' }}>
                         {isMenuOpen ? icons.cross() : icons.burger()}
                     </button>
                 </MobileMenu>
                 <Menu className={isMenuOpen ? 'open' : ''}>
                     <HomepageLink to="/features">Features</HomepageLink>
                     <HomepageLink to="/pricing">Pricing</HomepageLink>
-                    <HomepageLink to="/docs">Docs</HomepageLink>
+                    <div>
```
svenefftinge: I think this can be removed in favor of the combo box.

jankeromnes: @ChristinFrohne agrees that it's confusing to have the menu twice, so I'll remove it. Thanks for the feedback!

SharifPoetra:
The editor on mobile chrome very hard to control!

PR #16, opened 2019-03-20
Replace All: gitpod.io# → gitpod.io/#

jankeromnes:


jankeromnes:
Quick thought while editing the blog posts:

If Google finds the same content on [www.gitpod.io](https://www.gitpod.io) and [medium.com](https://medium.com), it's going to severely downgrade the SEO of the copy with the lower page rank (hint: it's pretty hard to beat medium.com's page rank).

Example: [This blog](https://www.quicksprout.com/2014/11/03/should-you-repost-your-blog-content-on-other-websites/) suddenly started losing 200K+ visitors every month because of reposts.

Two solutions to avoid this:
- Add `<link rel=canonical>` to reposts. Apparently [Medium can do that](https://help.medium.com/hc/en-us/articles/217991468-SEO-and-duplicate-content).
- Use only the other site as source of truth (i.e. don't import posts that are already on Medium)

PR #19, opened 2019-03-22
[mobile] index + features

jankeromnes:
Pixel 2

<details>
<summary>(old screenshot)</summary>
Pixel 2 full height
</details>

### DIFF
--- a/src/pages/features.tsx
+++ b/src/pages/features.tsx
@@ -7,7 +7,8 @@ import Logos from '../components/Logos';
 import GitGraph from '../components/GitGraph';
 import ModalVideo from 'react-modal-video';
 import 'react-modal-video/css/modal-video.min.css'
-import { colors } from '../styles/variables';
+import { colors, breakpoints } from '../styles/variables';
+import { getEmSize } from '../styles/mixins';
 import TerminalsImage from '../resources/terminals.png';
 import LanguageToolingImage from '../resources/language-tooling.png';
 import InlineCommentsImage from '../resources/inline-comments.png';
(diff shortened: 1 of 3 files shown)


jankeromnes:
Removed "Try It" Examples on mobile (it's less valuable there, and we have a bit too much content for a vertical layout already).

svenefftinge:
This part doesn't look right.
Screenshot 2019-03-22 at 17 06 34

jankeromnes:
> This part doesn't look right.

Yes, it needs some rework, hence the [WIP]. 🙂

I'll tweak how this section is created, by including all sections in the HTML (two with `display: none` on desktop, all in a list on mobile).

That way, images should also be pre-loaded by the browser on desktop. 👍

jankeromnes:
I did my best on the mobile "get started" items. Not 100% happy with it, and the code is a bit ugly, but it works:

Pixel 2 full page

jankeromnes:
Mobile features page (I just couldn't fix the `<table>`, I'll try to do that over the weekend):

Features on Pixel 2 (full page)

jankeromnes:
Ok, not the cleanest code, but the appearance of non-mobile pages remains almost exactly the same (with a few minor improvements in sizing -- I've updated the git-graph accordingly), and the pages now look much better on mobile.

Additional improvement: The "Get Started" images now get pre-loaded, that means that the page won't have to load a new image every time you switch to a new item.

And the pages now look almost perfect on mobile. We'll just have to fix the "Supported languages" table somehow (I tried to make it horizontally scrollable, but I didn't succeed).

PR #30, opened 2019-04-02
Minor fixes

jankeromnes:


src/pages/pricing.tsx
```diff
@@ -169,7 +169,7 @@ export default class PricingPage extends React.Component<{}, {}> {
                         </h3>
                         <div style={{ height: 3, width: 95, backgroundColor: colors.brand, marginBottom: 30 }} />
                         <p>
-                            Manage the subscriptions of teams at once.
+                            Manage the subscriptions of teams with a single invoice.
```
svenefftinge: I don't think we should stress the invoicing here.

jankeromnes: It was more about removing `at once`, which means `do that now, pronto`.

Please feel free to change the wording to your liking.

PR #33, opened 2019-04-02
Update prebuild docs to reflect latest state

32leaves:


svenefftinge:
I think that is because the link is not pointing to the market place, and the app doesn't do comments by default anymore.

PR #38, opened 2019-04-04
Tone down a few buttons and titles

jankeromnes:


svenefftinge:
The headers are no sometimes use title case and sometimes not.
Is there any reasoning behind when to use one or the other?

jankeromnes:
No specific reasoning. I just felt that "Titles That Are Very Long Get A Bit Unreadable In Title Case", and that "Do action" buttons are nice, but I didn't want to alter the look & feel of the index page too much, e.g. "Three Ways To Start Coding" is nice and readable, "No Setup", "Less Waiting" and "More Flow" work well with their respective icons, etc.

It's very subjective. Feel free to accept or refuse what feels right.

jankeromnes:
You can also just close this PR, I don't mind. 🙂 (I'm not 100% sure about "Try it now" vs "Try Now", but I do think "Request support" is a bit nicer than "Request Support", for example.)

svenefftinge:
I think it would be good to be consistent. So we decided on going without title case for buttons.
But for titles we usually use title case. it is a bot strange to make a few exceptions. That said I agree long titles get a bit shouty .

jankeromnes:
Let's revisit this later.

PR #40, opened 2019-04-05
[blog] Reduce font-weight just a notch

jankeromnes:
I think this tiny change makes the blog even more readable (and more like Medium).

Currently:

Screenshot 2019-04-05 at 12 28 09

With this change:

Screenshot 2019-04-05 at 12 28 14


jankeromnes:
`21px` was too big. Leaving at `20px` but `font-weight: 300` is more readable.

PR #41, opened 2019-04-05
Fix fragment links

geropl:


src/docs/70_Subscriptions.md
```diff
@@ -22,7 +22,7 @@ the end of the prior subscription.
 We may introduce other types of additional Gitpod hours (expiring or not) in the future. When you
 have different kinds of credits, the one with the earliest expiry date will be used first.
 
-You can see your current amount of Gitpod hours in the [dashboard](60_Dashboard.md#account-details)
+You can see your current amount of Gitpod hours in the [dashboard](/docs/60_Dashboard/#account-details)
```
geropl: I don't know how to make the `.md` link + fragment work, so I had to (?) use the absolute path.

jankeromnes: Sure, works for me, thanks!

PR #79, opened 2019-04-10
[docs] minor wording tweaks

jankeromnes:
Test:
- https://deploy-preview-79--gitpod-website.netlify.com/docs/40_Configuration/
- https://deploy-preview-79--gitpod-website.netlify.com/docs/41_Config_Gitpod_File/
- https://deploy-preview-79--gitpod-website.netlify.com/docs/42_Config_Docker/

### DIFF
--- a/src/docs/42_Config_Docker.md
+++ b/src/docs/42_Config_Docker.md
@@ -12,14 +12,12 @@ There are two ways to configure a custom Docker image in your `.gitpod.yml` file
     ```yaml
     image: node:alpine
     ```
-    The official Gitpod Docker images are hosted on [DockerHub](https://hub.docker.com/u/gitpod/).
+    The official Gitpod Docker images are hosted on [Docker Hub](https://hub.docker.com/u/gitpod/).
 * Reference a Dockerfile next to your `.gitpod.yml` file:
 
     ```yaml
     image:
-      file: docker/gitpod.Dockerfile
-      # Context is optional, defaults to an empty context
-      context: docker
+      file: .gitpod.dockerfile
     ```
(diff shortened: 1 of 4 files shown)


jankeromnes:
@geropl please take a look when you have time. Do these tweaks seem an improvement to you?

src/docs/42_Config_Docker.md
```diff
@@ -12,14 +12,12 @@ There are two ways to configure a custom Docker image in your `.gitpod.yml` file
...
     image: node:alpine
     ```
-    The official Gitpod Docker images are hosted on [DockerHub](https://hub.docker.com/u/gitpod/).
+    The official Gitpod Docker images are hosted on [Docker Hub](https://hub.docker.com/u/gitpod/).
 * Reference a Dockerfile next to your `.gitpod.yml` file:
 
     ```yaml
     image:
-      file: docker/gitpod.Dockerfile
-      # Context is optional, defaults to an empty context
```
geropl: I agree that the first example should be as simple as possible, but then we have to mention it somewhere I think. This was the one-size-fits-all solution.

jankeromnes: I agree that we need to mention some of these minor options somewhere. I'm planning to add a link to our schema, which could be a good first step.

The reason I removed it from here is that we'd like to encourage people to call their Dockerfiles `.gitpod.dockerfile`, which fits nicely next to `.gitpod.yml`. I'll try to find some way to also mention the `context:` option.

jankeromnes:
Cool, thanks a lot for the review Gero! 😄

PR #116, opened 2019-05-06
State requirement for Debian/Ubuntu based Docker images

csweichel:
in response to https://spectrum.chat/gitpod/general/custom-docker-image-ubuntu-debian-only~0498f5cf-2eb7-42c7-8373-c3945469e1bc, but should have been there all along.

### DIFF
--- a/src/docs/42_Config_Docker.md
+++ b/src/docs/42_Config_Docker.md
@@ -23,6 +23,8 @@ There are two ways to configure a custom Docker image in your `.gitpod.yml` file
 
 ## Creating Docker Images for Gitpod
 
+> Note: Gitpod only supports Debian/Ubuntu based images.
+
 A good starting point for creating custom Docker Images is the
 [`gitpod/workspace-full`](https://hub.docker.com/r/gitpod/workspace-full/) image. It already contains all the tools necessary to work with all languages Gitpod supports.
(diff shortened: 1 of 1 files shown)


src/docs/42_Config_Docker.md
```diff
@@ -23,6 +23,8 @@ There are two ways to configure a custom Docker image in your `.gitpod.yml` file
 
 ## Creating Docker Images for Gitpod
 
+> Note: Gitpod only supports Debian/Ubuntu based images.
```
geropl: We talked about that, but have we decided yet to drop alpine?

32leaves: indeed we haven't. It felt wrong to "announce" alpine support here considering we might be dropping it soon. It's always easy to "add"/announce a feature compared to removing support for one.

32leaves: and Alpine isn't working at the moment

svenefftinge: Should be: `Currently, Gitpod only ...`

JesterOrNot: Mention that it is currently being worked on and possibly send a link to the issue so more people can pitch in

JesterOrNot:
This was fixed in https://github.com/gitpod-io/website/pull/466

PR #144, opened 2019-05-27
Lighthouse score improvements

nisarhassan12:


src/pages/careers.tsx
```diff
@@ -241,7 +242,7 @@ const CareersPage: React.SFC<{}> = () => (
                 <h3>Jobs</h3>
                 <Job>
                     <h2>Gitpod Developer Advocate</h2>
-                    <p>As a Developer Advocate, you will connect with developers around the world, try out new frameworks and technologies, and share what you’ve learned with the world. You will craft great resources to help developers build with Gitpod, including sample projects, blog posts, and documentation. You’ll help grow our developer community, champion the interests of our users and translate their feedback into actionable product insights.
+                    <p>As a Developer Advocate, you will connect with developers around the world, try out new frameworks and technologies, and share what you’ve learned with the world. You will craft great resources to help developers build with Gitpod, including sample projects, blog posts, and documentation. You’ll help grow our developer community, champion the interests of our users and translate their feedback into actionable product insights.</p>
```
jankeromnes: Oops, thanks a lot for fixing my mistake!

src/pages/index.tsx
```diff
@@ -145,7 +146,7 @@ const WhyBox = styled.div`
     }
 `;
 
-const ResponsiveTweet = styled.div`
+const ResponsiveTweet = styled.blockquote`
```
jankeromnes: Oops, this creates a weird style in our tweets on the front page:

Screenshot 2019-05-27 at 15 48 59

Was that on purpose? (I find the fat left vertical bar a bit distracting, and preferred the previous tweet appearance).

src/styles/typography.ts
```diff
@@ -0,0 +1,15 @@
+import styled from '@emotion/styled';
+import {colors} from './variables';
+
+const UnderLine = styled.span`
+    &::after {
```
jankeromnes: Interesting, why did you make this a `::after` pseudo-element? (I'm not saying it's a bad idea, I'm just curious what makes this approach better).

src/components/ExampleBox.tsx
```diff
@@ -15,7 +15,7 @@ const ExampleBox: React.SFC<ExampleBoxProps> = (p) => (
     }}>
         <p style={{marginTop: 10, maxWidth: '90%', textAlign: 'center'}}>{p.description}</p>
         <img alt={`${p.title} Logo`} style={{ height: 100, maxWidth: 80, flexGrow: 2 }} src={p.image} />
-        <a href={p.link} target='_blank'>
+        <a href={p.link} target='_blank' rel="noreferrer">
```
jankeromnes: Ah, I remember that LightHouse suggests this. But here, we are linking from `www.gitpod.io` to `gitpod.io`, so I think there is no need to keep the referer secret. And maybe we'd like to actually know that the visit was referred from our website? If you agree, please remove this again.

src/components/Footer.tsx
```diff
@@ -72,7 +72,7 @@ const Footer: React.SFC<FooterProps> = () => {
                 <a href="https://twitter.com/gitpod"><SocialImg alt="Twitter Icon" src={twitter}/></a>
             </div>
             <div style={{ height: 40, marginTop: 15, display: 'flex', fontSize: 10, color: colors.fontColor1 }}>
-                Copyright © 2019&nbsp;<a href="https://typefox.io" target="_blank">TypeFox</a>&nbsp;| All Rights Reserved |&nbsp;<Link to="/imprint">Imprint</Link>&nbsp;|&nbsp;<Link to="/privacy">Privacy Policy</Link>&nbsp;|&nbsp;<Link to="/terms">Terms of Service</Link>
+                Copyright © 2019&nbsp;<a href="https://typefox.io" target="_blank" rel="noreferrer">TypeFox</a>&nbsp;| All Rights Reserved |&nbsp;<Link to="/imprint">Imprint</Link>&nbsp;|&nbsp;<Link to="/privacy">Privacy Policy</Link>&nbsp;|&nbsp;<Link to="/terms">Terms of Service</Link>
```
jankeromnes: Same, please remove this again (knowing about referrals between `typefox.io` and `gitpod.io` can be useful for us).

src/components/Header.tsx
```diff
@@ -166,7 +166,7 @@ export default class Header extends React.Component<HeaderProps, HeaderState> {
                     <HomepageLogo alt="Gitpod Logo" src={logo} />
                 </HomepageLink>
                 <MobileMenu role="navigation">
-                    {isMenuOpen ? null : <a href="https://gitpod.io/api/login">
+                    {isMenuOpen ? null : <a href="https://gitpod.io/api/login" rel="noreferrer">
```
jankeromnes: Please remove.

src/components/Header.tsx
```diff
@@ -178,8 +178,8 @@ export default class Header extends React.Component<HeaderProps, HeaderState> {
                     <HomepageLink to="/pricing">Pricing</HomepageLink>
                     <HomepageLink to="/docs">Docs</HomepageLink>
                     <HomepageLink to="/blog">Blog</HomepageLink>
-                    <HomepageA href="https://spectrum.chat/gitpod" target="_blank">Community</HomepageA>
-                    <a href="https://gitpod.io/api/login">
+                    <HomepageA href="https://spectrum.chat/gitpod" target="_blank" rel="noreferrer" >Community</HomepageA>
+                    <a href="https://gitpod.io/api/login" rel="noreferrer">
```
jankeromnes: Please remove.

src/pages/features.tsx
```diff
@@ -161,19 +161,19 @@ export default class FeaturesPage extends React.Component<{}, FeaturesPageState>
...
+                        button={<a href="https://github.com/marketplace/gitpod-io" target="_blank" rel="noreferrer">
                             <button className='primary'>GitHub Marketplace</button>
                         </a>}
                         visual={<img src={PrebuiltWorkspacesImage} alt="prebuilt workspaces" />}>
                         <p>Gitpod continuously builds your code and pull requests in the background to save your time.</p>
                     </Feature>
                     <Feature
                         title="VS Code"
-                        button={<a href="https://www.theia-ide.org" target="_blank">
+                        button={<a href="https://www.theia-ide.org" target="_blank" rel="noreferrer">
```
jankeromnes: Please remove.

src/pages/features.tsx
```diff
@@ -161,19 +161,19 @@ export default class FeaturesPage extends React.Component<{}, FeaturesPageState>
...
                             <button className='primary'>Visit theia-ide.org</button>
                         </a>}
                         visual={<img src={LanguageToolingImage} alt="language tooling" />} >
-                        <p>Gitpod is based on <a href="https://www.theia-ide.org" target="_blank">Theia</a>, the open-source VS Code-powered IDE optimized for developing in the cloud.
+                        <p>Gitpod is based on <a href="https://www.theia-ide.org" target="_blank" rel="noreferrer">Theia</a>, the open-source VS Code-powered IDE optimized for developing in the cloud.
```
jankeromnes: Please remove.

src/pages/features.tsx
```diff
@@ -353,7 +353,7 @@ export default class FeaturesPage extends React.Component<{}, FeaturesPageState>
                 <Container>
                     <div style={{ textAlign: 'center', marginTop: 150, marginBottom: 150 }}>
                         <h3 style={{ marginBottom: '0.7rem' }}>Try It Out</h3>
-                        <a href="https://gitpod.io/api/login">
+                        <a href="https://gitpod.io/api/login" rel="noreferrer">
```
jankeromnes: Please remove again.

src/templates/doc.tsx
```diff
@@ -167,18 +169,22 @@ function onSelectTopic(event: React.FormEvent<HTMLSelectElement>) {
...
-                {
-                    (m.subMenu || []).map(m =>
-                        <option key={m.path} value={m.path}>&nbsp;&nbsp;&nbsp;&nbsp;{m.title}</option>
-                    )
-                }
-            </>
-        })}
-    </select>
+    return (
+        <div>
```
jankeromnes: Ok, this file seems to only have white-space changes (i.e. the code or behavior were not actually modified), except here, where you're encapsulating the `<select>` into a `<div>`. Why?

jankeromnes:
Thanks again! 💯 I've addressed most of my own comments, so I think we can merge this now, but please take a look at those that aren't resolved yet.

src/pages/index.tsx
jankeromnes: I've reverted this particular change.

src/styles/typography.ts
nisarhassan12: Actually My intention was to remove that un-semantic div used as a blue underline and use a css class of .underlined and let css handle that for as its purely presentational and to keep the code dry i.e not having to repeat that `<div style={{height: 3, width: 95, background: ${colors.brand}, ...}}/>` everywhere we need that but could not find How to create a reusable class in emotion so ended up doing kind of the same thing.

src/templates/doc.tsx
nisarhassan12: I did that because I thought it would be better to have a label for the select input and later forgot about it.

Issue #141, opened 2019-05-24
Adding a "built on" section 


ChristinFrohne:
I think a "built on" section at the end of the git-graph on the main site would be good. 
It should contain logos from kuberneteres, Docker, Theia, VS Code, etc. 

It gives the user a fast overview about the tech behind Gitpod and increases trust

jankeromnes:
Maybe just above the "Explore Features", so that the call-to-action remains at the very end of the GitGraph?

ChristinFrohne:
I would say it fits the best just above "explore features"

PR #154, opened 2019-05-30
Created the built on section.

nisarhassan12:


jankeromnes:
@ChristinFrohne is this like what you had in mind for #141?

nisarhassan12:
Does this looks ok?
image

nisarhassan12:
image

jankeromnes:
Looks like this now:

Screenshot 2019-06-06 at 12 16 40

I really like it! Maybe there should be less space above and below the logos?

Also, regarding the texts, we've discussed that "Built on open-source projects" doesn't seem very valuable just on its own, and "gives you a very transparent code structure" doesn't make sense.

My suggestions were:

- "Built On Planet-Scale Technologies" (or "Built On The Shoulders Of Giants")
- "Gitpod is based on industry-proven open source software to bring you a seamless, safe, scalable and blazing fast development experience"

However, @ChristinFrohne found "planet-scale" a bit weird, and suggested "well-trusted", "well-known", or "reliable" instead.

jankeromnes:
Additional design thoughts:

- The letters `TS` in the TypeScript logo should be white
- Most of the blues are similar. What if we just used the same blue in every logo?
- And maybe make the Go logo blue as well?

ChristinFrohne:
I don't like to change official logos. It's a bit unfortunate that they have all almost the same blue. But I would either leave it this way or make them all white. I agree on the white TS and use the official blue GO logo as well. 
If planet-scale is commonly used, it's fine by me

svenefftinge:
- Theia be included (first?)
- the icons should have links to the repositories / websites.
- the git-graph dots are not properly aligned

Also, it is not totally clear to me what the purpose of the section is. Can someone elaborate a bit?

ChristinFrohne:
Its purpose is to give the user a fast overview about the tech behind Gitpod and to increase trust by showing the user brands which he's familiar with. It shall increase credibility. 
A side effect might be that the link structure is strengthened and the google pagerank increases, which leads to better SEO  
I noticed it on several other sites which I, of course, can't find right now. I can add them later on

jankeromnes:
Great work, thanks! Now looks like this:

Screenshot 2019-06-12 at 09 13 41

A few thoughts:
- "Explore Gitpod" section is much too close to the logos. Maybe move it down by `50px` or so.
- The final git-graph dot should be aligned with "Explore Gitpod" headline.
- The "S" of "TypeScript" should be uppercase.
- The word "Docker" should be centered under the Docker logo (should be `5px` left or so).

svenefftinge:
I like the revised wording. good job! :)
Anything left to do here?

ChristinFrohne:
Some comments: right now the dots are not aligned with the headlines, open-source is spelled once without `-` and why is "open-source" spelled with a capital letter? :)

jankeromnes:
@ChristinFrohne Good points, thanks! I'll address them shortly.

However, what do you mean by "the dots are not aligned"? They look fine to me:

Screenshot 2019-06-19 at 11 37 57

jankeromnes:
> They look fine to me:

(Ok, maybe they're not super exactly aligned, I'll try to make that better.)

> open-source is spelled once without `-`

I think "open source" is a noun (e.g. "I do open source"), while "open-source" is an adjective (e.g. "I'm an open-source developer").

> and why is "open-source" spelled with a capital letter? :)

Well, that doesn't make sense and looks weird. I wrote it that way, so I'll fix it. :)

ChristinFrohne:
> I think "open source" is a noun (e.g. "I do open source"), while "open-source" is an adjective (e.g. "I'm an open-source developer").

thanks, I didn't know that

That's how the dots look on my screen
Bildschirmfoto 2019-06-19 um 13 36 11

PR #194, opened 2019-06-21
Keyboard Navigation Fixes

nisarhassan12:
Made section "Three Ways to start coding" accessible for Keyboard Users by adding Keyboard support
image

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -206,6 +206,12 @@ const SectionBuiltOn = styled.section`
     }
 `
 
+const UnstyledButton = styled.button`
+    border: none;
+    padding: 0;
+    margin: 0;
+`
+
 export default class IndexPage extends React.Component<{}, IndexPageState> {
 
     protected getBrowserExtension(): string {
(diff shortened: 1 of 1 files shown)


svenefftinge:
I see blue focus boxes on the elements, but switching using cursor keys didn't work for me.
Screenshot 2019-06-22 at 20 17 06

nisarhassan12:
@svenefftinge I haven't added the support for cursor keys. I just made it to work with Tab(to move forwards ) and Shift + Tab(to move backwards). Is it necessary to support cursor keys as well?

svenefftinge:
>  Is it necessary to support cursor keys as well?
No it's not, I just assumed that was what you had done.

src/pages/index.tsx
```diff
@@ -206,6 +206,12 @@ const SectionBuiltOn = styled.section`
     }
 `
 
+const UnstyledButton = styled.button`
+    border: none;
+    padding: 0;
+    margin: 0;
```
jankeromnes: The blue outline @svenefftinge showed is actually visible in Chrome when you click on any of these items.

I don't think it looks good, so we should also "unstyle" the `outline` property:

```suggestion
    margin: 0;
    outline: 0;
```

PR #201, opened 2019-06-26
[blog] build a webhook for google assistant

anudeepreddy:


svenefftinge:
Hey @anudeepreddy,
I asked a friend of mine who is a Dev 🥑 @ google for feedback:

> Great article, looks good to me. A couple of suggestions:
>  - Obviously, people need to host this somewhere eventually, and - at least as far as I can tell - most people will use AWS or Firebase Cloud Functions (as opposed to pure Node/Express). Given that all our tutorials use Firebase, most developers will probably host their Action on Firebase.
>  - the key piece of the article is how to gitpodify your project, and I would suggest explaining the .gitpod.yml file in more detail 
>  - it also wouldn't hurt to actually build a Dialogflow agent - would make it more convincing. At the moment, this reads like a "Hey, we are Gitpod, let us show you how to use Gitpod to build an AoG fulfillment". I guess it should rather read like a "hey, I'm an AoG developer, and the turn-around times are so unbearably long. Here is a cool way how to speed up the turn-around times".

PR #218, opened 2019-07-07
moved images to CDN

anudeepreddy:
Hey @jankeromnes 
I moved the images to a CDN and as i can't server the images from the old url's i had to update the blog post here.


meysholdt:
thx!

PR #222, opened 2019-07-11
Created Github Student Developer Pack Page.

### LINKED ISSUE #219: GitHub Student Pack
URL: gitpod.io/github-student-developer-pack

**Links**
Try now = Try now section of the start screen 
Get student package = https://education.github.com/pack
Gitpod docs = https://www.gitpod.io/docs/10_Getting_Started/
Envelope icon = mailto:contact@gitpod.io?subject=Github%20Student%20Pack%20question
 
**Images**
[Github-student-package.zip](https://github.com/gitpod-io/website/files/3372408/Github-student-package.zip)

**Color**
the yellow line has the same color as the backpack #FDB02F 

**Mockup**
GitHub Student Package

nisarhassan12:
Fixes #219 
Here is What it looks like. @32leaves Please have a look! and tell if there is anything that can be improved : )
image
image


ChristinFrohne:
Looks great! 
Here some comments: 
**Website:**
- please have a bit less spacing between the GitHub Student Offer and the three boxes
- the sentence "it's great to support students changed to "We’re happy to be able to empower student developers participating in it." See last mockup
- the lines crossing out the price should have the same width as the characters
- it looks like the hand-shake-heart has a different glow than the other icons, please use for all icons the same setting as on the pricing page
- have the two lines a bit higher and some more space around the banner
- Is the font size in the banner the same size as "GitHub Student offer"? It would be great if it's not sticking out from the margins
 

**mobile:**
- more space between the backpack and the headline
- white line "always free for open-source" should be underneath the first box
- more space between the "gitpod docs" button and the next headline
- the footer doesn't look pretty but maybe that something to be adjust another time :)

32leaves:
image
I have two points where the website (mock and built page) differ from the rest of Gitpod:
- in Gitpod itself we call the "Students" plan "Personal" at the moment. We should call it the same way here.
- Both, the "Personal" and "Students Unlimited" plan are for non-commercial use only (see screenshot)

ChristinFrohne:
I think it's confusing to change the name of "unlimited" into "student unlimited" but "personal" stays the same. Also, the bag-icon is used to teaser the student subscription on the pricing site, but not used again. Anyway if we want to leave it like this for now, the icons should be the same as well.

32leaves:
> I think it's confusing to change the name of "unlimited" into "student unlimited" but "personal" stays the same. 

I second that - we'll use the correct name "Students" instead of "Personal" in this case. 

> Also, the bag-icon is used to teaser the student subscription on the pricing site, but not used again. Anyway if we want to leave it like this for now, the icons should be the same as well.

I have to check how much effort using the icons would add. But it would make sense to harmonize it all while we're at it.

PR #245, opened 2019-08-05
[docs] Automate /release-notes/all.json build

jankeromnes:
Fixes https://github.com/gitpod-io/gitpod/issues/714

### DIFF
--- a/build-release-notes.js
+++ b/build-release-notes.js
@@ -1,0 +1,16 @@
+'use strict'
+
+const cp = require('child_process');
+const fs = require('fs');
+
+process.chdir('static/release-notes/');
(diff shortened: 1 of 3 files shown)


jankeromnes:
... but I think this change might suddenly show the June release notes to many people.

Maybe let's merge this when we publish new release notes (e.g. in August).

svenefftinge:
The images are not shown properly, which is why I disabled the release notes.

jankeromnes:
> The images are not shown properly, which is why I disabled the release notes.

Thanks for the explanation. When we publish the next Release Notes, I'd like to try and fix image display, and then re-publish the June notes.

jankeromnes:
We have https://www.gitpod.io/docs/changelog/ now.

PR #264, opened 2019-10-08
Feedback contribute.dev

svenefftinge:
Hey @nisarhassan12, thanks for the post. I have created a PR so you can better see my proposed changes. The main proposal is to remove the complete section about how problematic setups usually are. I think this is too sidetracking. Instead I think it is good to keep the post light and positive just pointing out what contribute.dev is and how to use it.


svenefftinge:
cc @jankeromnes

nisarhassan12:
@svenefftinge Thanks! for the proposal, I agree! with it :slightly_smiling_face:  the post is now short, snappy and straight to the point.

PR #269, opened 2019-10-30
Fixed.  Image overlapping text issue:- Docs: **Getting Started**

### LINKED ISSUE #268: Docs: **Getting Started** . Image is overlapping text.
Currently, A Image in Docs: https://www.gitpod.io/docs/10_getting_started/ is overlapping the text. 

image

nisarhassan12:
Fixes #268 .

image

### DIFF
--- a/src/styles/normalize.ts
+++ b/src/styles/normalize.ts
@@ -341,4 +341,8 @@ export default `
         }
     }
   }
+
+  .gatsby-resp-image-wrapper {
+    margin-bottom: 4rem;
+  }
 `


svenefftinge:
@nisarhassan12 Thanks for taking care if this. It's great you are doing these kind of things on your own. I wonder however if this should not be generally taken care of in CSS. Couldn't this possibly happen in other places as well?

svenefftinge:
Maybe we can put some space below all all images. E.g. on this page the screenshots sit close to the text below as well https://www.gitpod.io/docs/54_git/

nisarhassan12:
Thanks! agree! We should make this change globally through out the `docs.

nisarhassan12:
Thanks! @svenefftinge I have made the change globally this is what it looks like: 

image

PR #271, opened 2019-10-31
Fixes Tweet collapsing issue.

nisarhassan12:
Currently, the last 2 tweets on the bottom right are collapsed and there is no spacing between them. This PR fixes that.

image

This is what it looks like know:

image

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -153,6 +153,12 @@ const ResponsiveTweet = styled.div`
         position: static;
         width: 106%;
     }
+
+    @-moz-document url-prefix(){
+        &.firefox--offset {
+            top: 622px !important;
+        }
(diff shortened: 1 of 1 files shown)


svenefftinge:
In chrome on OSX it looks good how it is:
Screenshot 2019-10-31 at 05 51 16

and gets worth with this change:
Screenshot 2019-10-31 at 05 51 37

nisarhassan12:
Thanks! @svenefftinge :slightly_smiling_face:  I have also tested this In Chrome it was looking worse with the new change. 

I found out that this collapsing problem occurs only in Firefox. So to fix this I wrote this

image

these styles will only get applied if the browser is Firefox. Now it looks good and consistent in both browsers.

PR #275, opened 2019-11-13
[blog] Using local services in Gitpod

32leaves:


src/blog/local-services-in-gitpod.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Using local services in Gitpod
+date: Thu Nov 13 2019 09:00:00 GMT+0000 (UTC)
+image: /local-services.jpg
+teaserImage: /local-services.jpg
+author: Christian Weichel
+---
+
+Some things just aren't exposed to the internet. Either becuase we do not want them out in the open, think some on-prem database, or because they cannot easily be made available like the Docker daemon on your laptop.
```
corneliusludmann: > Either becuase we do

typo: `because`

src/blog/local-services-in-gitpod.md
```diff
@@ -0,0 +1,77 @@
...
+  echo "token:  $token"
+  echo "remote: $(gp url 8090 | sed -e 's/https/wss/')"
+  ./inlets server --port 8080 --control-port=8090 --token="$token"
+  ```
+
+  Gitpod will show you notifications that there are new services on the ports `8080` and `8090`. Expose the port `8090` and leave the other one as is.
+
+2. run the inlets client on your local machine. The remote address and token are printed by the commands executed in step one.
+  ```
+  inlets client --remote=<see-above> --token=<see-above> --upstream=<local-addr>
```
corneliusludmann: How do I get the client on my local machine? Again by `curl -sLS https://get.inlets.dev | sh`? Or does it depends on my OS? Maybe a link to an install page?

corneliusludmann: Ah, I see below:
> On the client side (where I just assumed you already have inlets installed),

src/blog/local-services-in-gitpod.md
```diff
@@ -0,0 +1,77 @@
...
+  ```
+
+There's a lot going on there, so let's look at things in detail.
+In the workspace we first download `inlets` using its shell installer (we could have used `brew install inlets`, too, but that's a tad slower). Then we produce a token which prevents unauthorized access to our inlets server. Note that without tokens, the handshake performed by the inlets client and server fails.
+
+When starting the inlets server, we seperate the control port from the server port. We do this so that we can expose one port, but not the other. This way the control port is exposed to the internet (albeit secured with the access token), but the server port is not. This way we do not expose our local service globally, but only make it available within our Gitpod workspace.
+
+On the client side (where I just assumed you already have [inlets installed](https://github.com/inlets/inlets#install-the-cli)), we need to connect to the server we just started. To that end we use the corret remote URL (note the `wss://`) and token.
+What you choose as `--upstream` depends on which kind of service you want to expose.
```
corneliusludmann: I'm not sure what I should state here for `--upstream`. A locally running service, e.g. `--upstream localhost:3000` to make it available in Gitpod? Maybe you could give an example call here?

src/blog/local-services-in-gitpod.md
```diff
@@ -0,0 +1,77 @@
...
+
+Some things just aren't exposed to the internet. Either becuase we do not want them out in the open, think some on-prem database, or because they cannot easily be made available like the Docker daemon on your laptop.
+With your Gitpod workspace running in the cloud it does not have access to local services, which could render Gitpod unsuitable in those cases.
+
+There's a range of [different](https://developers.cloudflare.com/argo-tunnel/) [offerings](https://ngrok.com/) and [projects](http://serveo.net/) that aim to expose local services to the internet. However, they all do so indiscriminately. I.e. if one knows the URL to your now-public endpoint, they have access to your service.
+On top of that, they do incur additional cost if used often.
+
+## Inlets to the rescue
+
+![Inlets](local-services-in-gitpod/concept.png)
```
corneliusludmann: The white background does not look pretty neat with the dark website theme. Don't know if it could be improved without much effort.

src/blog/local-services-in-gitpod.md
32leaves: :+1:

src/blog/local-services-in-gitpod.md
32leaves: That's true. Unfortunately there's not so much I can do. I've rounded the corners of the white background a bit in an attempt to make it look nicer.

src/blog/local-services-in-gitpod.md
32leaves: :+1:

Issue #282, opened 2019-12-09
Thinner cookies consent


svenefftinge:
It is currently very bulky. I don't think it needs a button and can be much thinner.
Also a brighter background could help.

PR #284, opened 2019-12-09
improved cookie consent styling

svenefftinge:
Fixes gitpod-io/website#282


nisarhassan12:
Thanks! @svenefftinge  LGTM :) I have added a light off-white border to the top of the bar. I thought it would be nice to add a small subtle separator :slightly_smiling_face: 
image

With that change in place it looks like this
image

PR #291, opened 2019-12-10
add external icon on hover to external links in Footer.

nisarhassan12:
This is what it looks like: 
image


nisarhassan12:
@ChristinFrohne Should I also make the color of the external icon to blue ?

nisarhassan12:
> Does the icon have the same color as the gray font? It looks like it's darker. Can you adjust that? You're right a blue icon makes more sense on hover :) Please, make sure that you use the same style for "community" in the navigation

I have made this change.

ChristinFrohne:
Great! You can apply the same style to the navigation (community) which means to only show the icon on hover.

Issue #280, opened 2019-12-05
Adjustments for Website Relaunch


ChristinFrohne:
**General**

- [x] I'd like to try if the navigation works without the arrows(inspiration: https://circleci.com/) and have an adjusted drop-down-menu (see mockup)

- [x] please check the links, some are not correct, e.g. Choose your subscription should link to each table underneath

- [x] spelling mistakes (check for GitLab, GitHub, Trial, etc.)

- [x] let's change the button order in "Explore Gitpod": First "See Features" /"See Pricing" in blue and then "See Blog" in white

- [x] I think the border of the white button is too thin, can you try 2px? Can you make the buttons on the start screen a bit smaller?

**Home** 

- [x] Why Gitpod: If no hover is active, please display the initial graphic

- [x] Move Integration section up (also see updated text)

- [x] Align the first tweet with the headline, following tweets can scroll to the side like now

- [x] I updated the "Get Started" Section to make the form more recognizable. Does it work? What do you think?
<img width="964" alt="Bildschirmfoto 2019-12-05 um 12 09 45" src="https://user-images.githubusercontent.com/46345125/70230212-285d8580-1758-11ea-9fce-d8cd5750160e.png">

**Features**

- [x] could you add an arrow that scrolls with one and that can be clicked to get back to the overview?

@nisarhassan12 I also added a contact form and a 404 page to the mockup

ChristinFrohne:
More Feedback: 
- [ ] Please add more tweets (e.g. https://twitter.com/LastZactionHero/status/1188910483698372608, https://twitter.com/CuriousNikhyl/status/1117141675745402881, https://twitter.com/YourETHGirl/status/1120015913024139265, https://twitter.com/hughdurkin/status/1102215129696010240, https://twitter.com/YourETHGirl/status/1120015913024139265, https://twitter.com/rocksetta/status/1134511186286768129)

- [ ] I like the new navigation, but can you make the drop-down without the animation? (see https://circleci.com/ for inspiration) and use the bubble-box?

- [ ] Graphic Ready-to-code: Can you make sure that the graphics are exactly aligned on each other? Currently, it's jumping around. 

- [ ] Vendor, Recruiting, Education: should have less spacing between the boxes, the quote and the image

- [ ] education: "are you a student" is too big. See mockup for the recent design

- [ ] add an icon for "community" to show that an external site will open up

- [ ] Buttons on the start screen and navigation are too big. Can you have less padding around? 

[...]

PR #292, opened 2019-12-11
Update code review feature.

nisarhassan12:
This is what it looks like: 
image

PR #293, opened 2019-12-11
Features small updates/fixes.

nisarhassan12:
This PR addresses the following from #280 
image


nisarhassan12:
> The code review section still looks the same here. But I think you fixed that in another PR right?

Yes! that has been updated in this PR https://github.com/gitpod-io/website/pull/292.

Issue #286, opened 2019-12-09
Mobile view - layout adjustments


ChristinFrohne:
**Navigation:**

- [x] Less spacing at the top

- [x] smaller exit icon

- [x] Adjust each navigation point in the center of the lines

- [x] Add drop down menu
 Current view:
<img width="497" alt="Bildschirmfoto 2019-12-09 um 15 59 12" src="https://user-images.githubusercontent.com/46345125/70446328-359eab00-1a9d-11ea-87d4-b2ccc5e80e1a.png">

My suggestions:

<img width="284" alt="Bildschirmfoto 2019-12-09 um 16 05 39" src="https://user-images.githubusercontent.com/46345125/70446676-c37a9600-1a9d-11ea-83a0-0ee63818a82f.png">

**Home:**

- [x] Hover for Ready-to-code dev environments should also work on click/touch

- [x] "Seamless GitHub /GitLab integration" and "Full Dev environments": have the image first and then the text

- [x] "Don't just take our word" headline should have less spacing to the tweets

- [x] Prefix URL bar is just half shown

- [x] Add a horizontal slide to choose between the different programming languages

- [x] "Giving back to open-source" Make the white box bigger and have the screenshot inside the box

**Features:**

- [x] smaller overview boxes (vs code extensions, prebuilt workspaces, etc.)

- [x] Give the images in each box a bit more space to the top and a bit less space to each headline

**Enterprise, Vendor, Recruiting, Education**

- [x] quotation marks should have some space on the sides

ChristinFrohne:
- [ ] Tweets are overlapping on iOS
![IMG_4172](https://user-images.githubusercontent.com/46345125/70696270-a45d4d80-1cc3-11ea-8d1c-989e263aad2a.PNG)

PR #298, opened 2019-12-11
Nav updates

nisarhassan12:
This PR addresses things specified in #286 related to the Nav. 

image

this is what it looks like now: 
image


nisarhassan12:
@ChristinFrohne The drop-down is already like that please test on a Real Mobile Device. The hover support is for PCs(Desktops & Laptops) as you asked earlier do you want me to remove the hover support completely?
image

ChristinFrohne:
@nisarhassan12 okay, it's most important that it works that way on mobile devices. But I think that the hover doesn't make sense in the small menu generally. It's great though for the normal view. 
But we have more important things to work on, so we can leave it like this for now :)

nisarhassan12:
> @nisarhassan12 okay, it's most important that it works that way on mobile devices. But I think that the hover doesn't make sense in the small menu generally. It's great though for the normal view.
> But we have more important things to work on, so we can leave it like this for now :)

Thanks! I have made the suggested changes..

PR #301, opened 2019-12-12
features page updates mobile

nisarhassan12:
This PR addresses these things as specified in #286 
image


nisarhassan12:
Thanks! @ChristinFrohne it now fixed this is what it looks like now:
image

PR #313, opened 2019-12-16
[pricing] update the layout in accordance to the new mock-up.

### LINKED ISSUE #281: Layout Pricing
What do you think about this layout? The aim was to visualize that the Personal, Professional and Unlimited Plan is based on the open-source plan and to highlight more the free open-source plan. @svenefftinge @nisarhassan12 

Pricing

nisarhassan12:
Fixes #281 

image


nisarhassan12:
image

Those question marks can be useless on mobile because there is no hover support on touch devices. We also can't try click because of the question mark icon is too small thus hard to touch/tap with a finger. Small touch targets are not good for accessibility( for some people they can be as hard as threading a needle).

@jankeromnes @ChristinFrohne If the information that appears while hovering them is important maybe we can try something else :slightly_smiling_face:?

jankeromnes:
> Those question marks can be useless on mobile because there is no hover support on touch devices. We also can't try click because of the question mark icon is too small thus hard to touch/tap with a finger. Small touch targets are not good for accessibility( for some people they can be as hard as threading a needle).

You're making great points here.

- I'd suggest making the `?` icon a bit bigger, so that on mobile you can click/tap it reasonably easily to view the hover text.
- If that doesn't work, maybe we should show the small text directly on the page, and not hide it between interactive JS widgets.

ChristinFrohne:
I agree. It would be great if you can make the icons bigger on mobile view only. Showing its content directly might distract the user from the other content. 
Here a quick mockup that should work.
Bildschirmfoto 2019-12-16 um 14 07 51
Here you can also read more about it: https://uxmovement.com/mobile/how-to-display-tooltips-on-mobile-forms/

Issue #315, opened 2019-12-16
Layout changes (Summary)


ChristinFrohne:
Let's summarize what has to be done in this issue, so nothing gets left behind. Please add tasks if something slipped my mind. 

- [x] Navigation drop down: please use the suggested speech bubble and delete the animation. Just let the bubble appear. #280 

- [x] Add light blue hover for all links 

- [x] Can you try out using the semi bold font for the buttons to make them stick out a bit more 

- [x] have less spacing above the zick zack line 

- [x] update sand glass icon (see mockup)

- [x] Trusted by: Do all logos have the same gray? It doesn't seem like it (#9f9f9f)

- [ ] Adjust "ready-to-code" graphic: align the images perfectly and have the Gitpod logo in front of the "With Gitpod graphic" as well. #280 

- [ ] Update Theia image
![Theia-Ide-icon](https://user-images.githubusercontent.com/46345125/70893402-4eeaae80-1feb-11ea-885d-eec740a43c0a.png)

- [x] Get started URL bar: Have the focus active be default. I'll work on another mockup to highlight the section more

- [ ] features: Add hover for the boxes in the features' overview #280 

- [x] features: have the illustrated images a bit smaller (share running workspace, snapshot, parallel workspaces)

- [x] features:  when the background changes to gray the border of the boxes should be white (see mockup)

- [x] pricing: see #313 

- [ ] "Contact Sales" should link to the contact form with a preset subject #280 

- [ ] Contact form: Subject is missing and font color in the form should be the same #280

- [ ] I guess "schedule a call" should link to calendly (https://calendly.com/gitpod/onboarding?back=1&month=2019-12) Maybe we should create another event or update the messaging so it fits to enterprise, education, vendor and recruiting as well. @svenefftinge? #280 

- [x] Education: delete "Are you a student - tell your teachers about it" #280 

- [x] documentation: add blue hover the edit-icon and the arrows to get to the next page

- [x] Update pricing table (add hover and question marks only if there is time left) #280
<img width="707" alt="Bildschirmfoto 2019-12-12 um 10 44 32" src="https://user-images.githubusercontent.com/46345125/70895820-e0f4b600-1fef-11ea-90e6-5848304d6920.png">

- [x] afterwards please start exchanging all .png galaxy images with .svg images. Beginning with the home screen, followed by the features. And let's see how far you get until Wednesday :) 

**If you have time left before Wednesday:**
[...]

PR #324, opened 2019-12-17
Layout Changes: #1

nisarhassan12:
This PR addresses the following things mentioned in #315 

image


nisarhassan12:
> Nice!
> Just one thing: The spacing between the logos isn't even. The space between freecopecamp and code institute is bigger. Please make it even. You can use the total width of the box underneath

image

Thanks! @ChristinFrohne that's because the logos are different in heights & widths. So if I were to replace code institute logo with freecodecamp it will look even.

image

PR #330, opened 2019-12-17
[docs] Fix menu regression

32leaves:
We had some slight regression of the docs menu content w.r.t. to self-hosted. This PR fixes that.

### DIFF
--- a/src/docs/menu.ts
+++ b/src/docs/menu.ts
@@ -137,23 +137,15 @@ export const MENU: MenuEntry[] = [
         "self-hosted/latest/self-hosted",
         [
             M(
-                "Prepare Install",
-                "self-hosted/latest/install/01_prepare_install/"
+                "Prerequisites",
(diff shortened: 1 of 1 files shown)


src/docs/menu.ts
```diff
@@ -137,23 +137,15 @@ export const MENU: MenuEntry[] = [
         "self-hosted/latest/self-hosted",
         [
             M(
-                "Prepare Install",
-                "self-hosted/latest/install/01_prepare_install/"
+                "Prerequisites",
+                "self-hosted/latest/install/01_prepare_installation/"
             ),M(
-                "Install on Kubernetes",
+                "Install on Vanilla Kubernetes",
```
meysholdt:
```suggestion
                "Install on Generic Kubernetes",
```

meysholdt: It's Vanilla in other places, too, so ignore this change suggestion.

PR #331, opened 2019-12-18
Get started input updates

nisarhassan12:
This PR address this https://github.com/gitpod-io/website/issues/315#issuecomment-566036797 .
This is what it looks like now: 
image

PR #341, opened 2019-12-18
[index] input add autoFocus + add galaxy bg to the prefix in the text.

nisarhassan12:
This PR address: https://github.com/gitpod-io/website/pull/331#pullrequestreview-333785251

### DIFF
--- a/src/components/GetStarted.tsx
+++ b/src/components/GetStarted.tsx
@@ -292,7 +292,6 @@ class GetStarted extends React.Component {
                                         defaultValue={val}
                                         onChange={this.handleChange}
                                         onKeyPress={this.handleKeyPress}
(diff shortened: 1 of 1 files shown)


svenefftinge:
It now always jumps to the getting started section.

Also the `gitpod.io/#` is replaced with a slice of the galaxy bg.
Screenshot 2019-12-18 at 18 09 13

nisarhassan12:
Thanks! This is what it looks like in Firefox:

image

I think I should make it my habit to always check on Chrome also.

nisarhassan12:
I have fixed the issues.

svenefftinge:
Thanks, Nisar. It doesn't set the focus but I think it's not so important.

PR #350, opened 2019-12-19
[self-hosted] Better illustrative install steps?

jankeromnes:
That way, the commands actually "work" (will likely tell you that you need to do more stuff, but at least the example commands are valid and you can iterate from there).

Also, I think it's useful to tip off users about `values.yaml` being an important file (otherwise, you don't know any configuration files that you can change).

### DIFF
--- a/src/blog/self-host-your-gitpod.md
+++ b/src/blog/self-host-your-gitpod.md
@@ -47,9 +47,9 @@ All it needs to get it running on your infrastructure is to have a domain name a
 
 1. In your terminal, run: `git clone https://github.com/gitpod-io/self-hosted`
 
-2. Adjust the configuration from the repository to your needs.
+2. Adjust configuration files like `values.yaml` to your needs
 
-3. Run `helm install -f ... gitpod`
+3. Run `helm install -f values.yaml gitpod .`
 
 [The setup for vanilla Kubernetes](/docs/self-hosted/latest/install/10_install_on_kubernetes/) will work most Kubernetes clusters.
 For Google Cloud Platform (GCP) we offer a [script that automates the install](/docs/self-hosted/latest/install/11_install_on_gcp_script/).
--- a/src/pages/self-hosted.tsx
+++ b/src/pages/self-hosted.tsx
@@ -123,18 +123,18 @@ const SelfHostedPage: React.SFC<{}> = () => (
                         <ol className="install__steps">
                             <li className="install__step">
                                 <span>1.</span>
-                                <StepP>In your terminal run:</StepP>
+                                <StepP>In your terminal, run:</StepP>
                                 <Terminal>
                                     git clone <a href="https://github.com/gitpod-io/self-hosted" target="_blank">https://github.com/gitpod-io/self-hosted</a>
                                 </Terminal>
                             </li>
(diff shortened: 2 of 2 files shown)


src/blog/self-host-your-gitpod.md
```diff
@@ -47,9 +47,9 @@ All it needs to get it running on your infrastructure is to have a domain name a
 
 1. In your terminal, run: `git clone https://github.com/gitpod-io/self-hosted`
 
-2. Adjust the configuration from the repository to your needs.
+2. Adjust configuration files like `values.yaml` to your needs
```
32leaves: This is to diminutive ... `values.yaml` alone never suffices. You either need the registry config or https.

src/pages/self-hosted.tsx
```diff
@@ -123,18 +123,18 @@ const SelfHostedPage: React.SFC<{}> = () => (
...
                                 <span>1.</span>
-                                <StepP>In your terminal run:</StepP>
+                                <StepP>In your terminal, run:</StepP>
                                 <Terminal>
                                     git clone <a href="https://github.com/gitpod-io/self-hosted" target="_blank">https://github.com/gitpod-io/self-hosted</a>
                                 </Terminal>
                             </li>
                             <li className="install__step">
-                                <span>2.</span><StepP>Adjust the configuration from the repository to your needs.</StepP>
+                                <span>2.</span><StepP>Adjust configuration files like `values.yaml` to your needs</StepP>
```
32leaves: see above

src/blog/self-host-your-gitpod.md
```diff
@@ -47,9 +47,9 @@ All it needs to get it running on your infrastructure is to have a domain name a
...
 
-3. Run `helm install -f ... gitpod`
+3. Run `helm install -f values.yaml gitpod .`
```
32leaves: Not: for someone who knows helm it feelds odd to explicitly list `values.yaml` using `-f`

32leaves: maybe `helm install ... gitpod` works. It doesn't run OOTB, but is easy to iterate on and makes it clear that more needs to be added here.

src/blog/self-host-your-gitpod.md
jankeromnes: I was able to start a Gitpod install without these. Maybe this is better left for the details in the further docs?

I was going for a minimal set of example instructions that do something interesting.

src/blog/self-host-your-gitpod.md
jankeromnes: Ah, even better if that's the default file. Makes the example command shorter. :)

jankeromnes: It says this:

```
$ helm install ... gitpod
Error: failed to download "gitpod" (hint: running `helm repo update` may help)
```

The suggestion didn't fix it.

jankeromnes: And the current example is not even a valid command:

```
$ helm install -f ... gitpod
Error: must either provide a name or specify --generate-name
```

jankeromnes:
I give up.

Issue #379, opened 2020-01-03
FreeCodeCamp logo is too small on mobile


jankeromnes:
On mobile, the FreeCodeCamp logo on the front page is too small. It's a shame, because it's super cool that we're partnering with FreeCodeCamp. 🙂

![Screenshot_20200103-183550](https://user-images.githubusercontent.com/599268/71738869-30443e80-2e58-11ea-99db-137ca67245ac.png)

It could be much bigger (e.g. by allow it's width to be larger on mobile). I think it's fine to special-case this logo, as the others already look good.

PR #382, opened 2020-01-06
[trusted-by] make FCC logo bigger for mobile devices.

nisarhassan12:
Fixes gitpod-io/website#379

image

### DIFF
--- a/src/components/TrustedBy.tsx
+++ b/src/components/TrustedBy.tsx
@@ -39,6 +39,14 @@ const StyledTrustedBy = styled.section`
             grid-template-columns: repeat(1, 1fr);
         }
     }
+
+    .fcc {
+        transform: scale(1.05);
+
+        @media(max-width: ${sizes.breakpoints.sm}) {
+            transform: scale(1.3);
+        }
+    }
 `
 
 const StyledBrandImage = styled.img<{transform?: string}>`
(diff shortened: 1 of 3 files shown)


JesterOrNot:
Nice one!

src/pages/enterprise.tsx
```diff
@@ -93,7 +93,7 @@ const EnterprisePage: React.SFC<{}> = () => (
                     alt: 'freeCodeCamp.org',
                     url: 'https://www.freecodecamp.org/',
                     svg: FreeCodeCamp,
-                    transform: 'scaleY(1.3)'
```
jankeromnes: Aaah, that's why the logo was stretched vertically! 😲 Thanks for removing it.

src/components/TrustedBy.tsx
```diff
@@ -39,6 +39,14 @@ const StyledTrustedBy = styled.section`
             grid-template-columns: repeat(1, 1fr);
         }
     }
+
+    .fcc {
+        transform: scaleY(1.3);
```
jankeromnes: Please always preserve aspect ratios for logos (i.e. never just `scaleX()` or `scaleY()`). If you need it bigger or smaller, please use `scale()`.

```suggestion
        transform: scale(1.3);
```

(Or remove this line if you didn't want the logo wider.)

Issue #383, opened 2020-01-06
Document Languages Supported In Gitpod


JesterOrNot:
# Please reference this issue in related PRs
### Key: 
* Just checked off = Not done but someone is working on it
* Checked off + Strike-through  = Done!
* Not Checked or Strike-through but Bold and Italics = Started but needs work (Will specify what needs work)
* Else = Up for grabs let me know so I can mark it as such

- [ ] JavaScript & TypeScript (but let's skip these for now -- I think we should make framework-specific pages for these)
- [ ] ***Python & Jypyter Notebook (Needs work)***
- [ ] HTML & CSS
- [ ] ***Java (Needs debugging info)***
- [ ] Dockerfile
- [ ] C++ & C
- [x] Go
- [ ] Shell (Bash)
- [ ] Ruby
- [ ] PHP
- [ ] Vue
- [ ] Scala
- [ ] Rust
- [ ] C# / .NET
- [ ] Dart
- [ ] Xtend
- [ ] R
- [ ] Elm
- [ ] Kotlin

For Each language
- Create a dedicated "Languages & Frameworks" page, briefly explaining what you need in your .gitpod.yml and/or .gitpod.Dockerfile
- List which VS Code extensions you should install & briefly show what each extension does (e.g. with screenshots)
- Link to a great example repository (ideally, a gitpodified project that is already popular/famous in that language, or a gitpodified fork)
- If relevant, explain how to configure debugging, e.g. how to create a debug configuration in your .theia/ or .vscode/ directory
- If our GitHub issues or Discourse community contain helpful infos about the language, add links to search our issues / our Discourse
- If possible, update the .gitpod.yml-inferrer with an automatic setup for that language
- If the example repository is really cool, we should probably also write a short blog post about its setup, and post it to Gitpod blog (+ cross-post to Dev.to)

JesterOrNot:
Here are a few wish list items
* Electron
* Flutter
* Tensorflow

JesterOrNot:
Ordering reference

> Please try to order languages by popularity in Gitpod (with a few exceptions, e.g. when a guide is really well written).
> 
> ### Gitpod Top Languages (2020-01-02)
> 1. JavaScript
> 2. Python
> 3. TypeScript
> 4. HTML
> 5. Java
> 6. Dockerfile
> 7. CSS
> 8. C++
> 9. Go
> 10. Shell
> 11. Ruby
> 12. PHP
> 13. Vue
> 14. Jupyter Notebook
[...]

PR #386, opened 2020-01-07
[docs] Add HTML to languages-and-frameworks.md

JesterOrNot:
This adds a article to the docs about how to configure Gitpod for HTML and CSS
see #383


src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
+# HTML
+
+Looking to try web-development in Gitpod? Well your in luck Gitpod's IDE - Theia - was made with web-development in mind, but depending on your needs you may want to customize this experience further!
```
jankeromnes: Typos:

```suggestion
Looking to try web development in Gitpod? Well you're in luck: Gitpod was made with web development in mind. And, depending on your needs, you may want to customize this experience further!
```

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Workspace Configuration
+
+### VSCode Extensions
+
+#### [HTML-Snippets](https://marketplace.visualstudio.com/items?itemName=abusaidm.html-snippets)
+This extension adds many usful snippets for productivity, see for yourself.
```
jankeromnes: Typos:

```suggestion
This extension adds many useful snippets for productivity. See for yourself:
```

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+![Example](images/HTML-Snippets-Example.gif)
+#### [HTML CSS Support](https://marketplace.visualstudio.com/items?itemName=ecmel.vscode-html-css)
+This will add great auto-complete for your project including color comlpetion!
```
jankeromnes: Typos:

```suggestion
This will add great auto-completion for your project, including color-completion!
```

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+![HTML-CSS-Support](images/HTML_CSS_Extension.png)
+
+### Pre-Builds
```
jankeromnes: As a developer who is new to Gitpod, I don't think this section title gives me any useful information.

Maybe `## Configuration files` instead? Or directly `## .gitpod.yml`?

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+With Gitpod you can have your project built before you open your repository, no more waiting for all of those Node/Yarn dependencies to be built! Let's get started with building your project to do this we need a [`.gitpod.yml`](https://www.gitpod.io/docs/41_config_gitpod_file/) file. Here is an example of what your start tasks should look like for a typical NPM project
```
jankeromnes: I suggest using more punctuation:

```suggestion
With Gitpod you can have your project built before you open your repository, no more waiting for all of those Node/Yarn dependencies to be built! Let's get started with building your project: To do this we need a [`.gitpod.yml`](https://www.gitpod.io/docs/41_config_gitpod_file/) file. Here is an example of what your start tasks should look like for a typical NPM project:
```

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+    command: npm start
+```
+
+For Yarn
+```YAML
+tasks:
+  - init: yarn && yarn build
+    command: yarn run
+```
+These should prebuild your workspace whenever you push a change!
```
jankeromnes: That's not exactly true. By default, this will only build your workspace whenever you create a new one.

To build things _before_ a workspace is opened (i.e. prebuilds), you need to install the [Gitpod app](https://github.com/marketplace/gitpod-io/) first. You'll also need to make sure that it's enabled for your workspace.

Additionally, keep in mind that only the `init:` instructions are prebuilt. The `command:` is supposed to be a quick command that starts an app interactively, which is why is doesn't fit well with prebuilds conceptually.

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+
+Next lets talk about previewing your setup without leaving the editor. To do this we must configure ports say your server listens on port 3000 in your [`.gitpod.yml`](https://www.gitpod.io/docs/41_config_gitpod_file/) file add this
```
jankeromnes: Typos:

```suggestion
Next let's talk about previewing your HTML pages without leaving the editor. To do this we must configure ports say your server listens on port 3000 in your [`.gitpod.yml`](https://www.gitpod.io/docs/41_config_gitpod_file/) file add this
```

jankeromnes: I think there are several possible scenarios here, when you want to preview HTML files.

1) You want to live-preview just a single HTML file. For this, you can click on the small Live Preview icon on the top-right corner of your HTML editor, and this will open an HTML live preview.

2) You want to preview several files, e.g. an HTML file that imports external scripts, CSS files, images, etc. For this, you can simply:
  a) run `python3 -m http.server 3000` in your workspace directory
  b) use the Gitpod notification to open port 3000 in a Preview
  c) open the HTML file you want to preview

3) You already have a server (e.g. via NPM or Yarn or something else like Ruby-on-Rails etc). Then you can configure your server port to automatically open a Preview like you explained.

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+# Declare ports section
+ports:
+    # Expose port 3000
+  - port: 3000
+    # Open preview after port opens
+    onOpen: open-preview
+```
+### Try It!
+
+Want to try it out? And see a minimal example in action? Great, then open it in Gitpod!
```
jankeromnes: Maybe a little bit less insistent on the usage of Gitpod? 😅 

```suggestion
Want to try it out? And see a minimal example in action? Great, then you can open this in Gitpod:
```

src/docs/html_in_gitpod.md
```diff
@@ -0,0 +1,50 @@
...
+
+[![JesterOrNot/Gitpod-Web-Development-Example](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/JesterOrNot/Gitpod-Web-Development-Example)
+## External Resources
+* [Start Tasks](https://www.gitpod.io/docs/44_config_start_tasks/)
+* [Exposing Ports](https://www.gitpod.io/docs/43_config_ports/)
+* [Theia](https://www.gitpod.io/docs/50_ide/)
+* [Prebuilds](https://www.gitpod.io/docs/46_prebuilds/)
```
jankeromnes: These external links feel a bit random. But I'm not opposed to having them.

src/docs/html_in_gitpod.md
JesterOrNot: Sorry my bad!

src/docs/html_in_gitpod.md
JesterOrNot: Added descriptions

src/docs/html-in-gitpod.md
```diff
@@ -0,0 +1,65 @@
+# HTML
+
+Looking to try web development in Gitpod? Well you're in luck: Gitpod was made with web development in mind. And, depending on your needs, you may want to customize this experience further!
+
+## Workspace Configuration
+
+### VSCode Extensions
```
nisarhassan12: Maybe there is no need of mentioning VSCode extensions here as Gitpod by default comes with **Emmet** which has great HTML and CSS support and I do not see any clear benefit of adding **HTML-Snippets** extension here. @JesterOrNot am i missing something :slightly_smiling_face: ?

nisarhassan12:
Thanks! for taking care of this :slightly_smiling_face: I agree with @jankeromnes It would make sense to have separate individuals guides, one for HTML and CSS and the other for JavaScript and TypeScript.

In this one, I think we don't have to mention npm or yarn tasks, ports and stuff like that instead we can mention about Emmet default HTML, CSS support and Live Preview that reflects changes on the fly
.
image

src/docs/html-in-gitpod.md
JesterOrNot: Because without it there is almost nothing to configure, and if that is the case we can probably move this whole article to theia tips and tricks!

JesterOrNot:
Ahh! Wrong button :( Will redo PR

PR #397, opened 2020-01-16
Fixes #394

### LINKED ISSUE #394: [accessibility] make the dropdown & nav items appear on focus.
Currently, In order for the Keyboard users to use the dropdown or the menu when the items are hidden for small devices they must press enter to view their content but this is not how it's supposed to work both should reveal their contents when they get focus state.

nisarhassan12:
Fixes #394


nisarhassan12:
> The navigation bar disappears after clicking through it
> 
> Bildschirmfoto 2020-01-22 um 09 29 12

@ChristinFrohne Thanks for the review great catch! I have fixed  it.

PR #428, opened 2020-01-24
prebuilds: Add method to trigger prebuild using URI

Kreyren:
Signed-off-by: Jacob Hrbek <kreyren@rixotstudio.cz>

### DIFF
--- a/src/docs/prebuilds.md
+++ b/src/docs/prebuilds.md
@@ -59,6 +59,12 @@ You can enable prebuilds also for all branches and for pull requests from forks.
 ### GitHub integration
 Once the GitHub app is installed Gitpod can add helpful annotations to your pull requests.
 
+### Manual execution of prebuild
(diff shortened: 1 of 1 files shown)


src/docs/prebuilds.md
```diff
@@ -59,6 +59,12 @@ You can enable prebuilds also for all branches and for pull requests from forks.
 ### GitHub integration
 Once the GitHub app is installed Gitpod can add helpful annotations to your pull requests.
 
+### Manual execution of prebuild
+Alternatively and independantly from configuration above it is possible to trigger the prebuild using
+```
+https://gitpod.io/#prebuild/https://github.com/ORG/REPO
```
csweichel: You can use any context URL here, much like you would without the prebuild prefix.
E.g. `https://gitpod.io/#prebuild/https://github.com/csweichel/bel/tree/master` would also work.

PR #433, opened 2020-01-24
Removed dashboard term from docs.

jbicker:
Dashboard page got removed from docs and respective parts (links and texts) changed to different terms.


src/docs/sharing-and-collaboration.md
```diff
@@ -15,8 +15,8 @@ and other UI state are currently not shared.
 
 ### How To Share and Unshare a Running Workspace
 
-Running workspaces can be shared from the [dashboard](/docs/dashboard/) or from within the IDE. In the IDE,
-right-click on the user's avatar and choose
+Running workspaces can be shared from within the IDE. 
+Click on the user's avatar and choose
```
jankeromnes: Oops, good catch!

src/docs/subscriptions.md
```diff
@@ -22,8 +22,8 @@ the end of the prior subscription.
 We may introduce other types of additional Gitpod hours (expiring or not) in the future. When you
 have different kinds of credits, the one with the earliest expiry date will be used first.
 
-You can see your current amount of Gitpod hours in the [dashboard](/docs/dashboard/#account-details)
-or in the IDE (_Help > Account Details_). You will be warned multiple times when you are getting
+You can see your current amount of Gitpod hours in the [usage page](/usage)
```
jankeromnes: I don't think this relative link will work -- did you mean to link to https://gitpod.io/usage/ ? (Different domain than the docs)

src/docs/subscriptions.md
```diff
@@ -22,8 +22,8 @@ the end of the prior subscription.
...
+or in the IDE by clicking your avatar and choosing _Account_. You will be warned multiple times when you are getting
 close to running out of Gitpod hours. As soon as your credits are used up, your workspaces will be
-closed automatically. You can still use the dashboard, but you will not be able to start any
+closed automatically. You can still use the account page, but you will not be able to start any
```
jankeromnes: Not too happy about the "account page" term here (because that doesn't include the multiple pages "workspaces", "usage", "subscription", "env variables", etc) but I don't see a much better alternative.

Maybe "account pages" plural? :/ Or "You can still manage your account"?

src/docs/subscriptions.md
jbicker: Link adapted

src/docs/subscriptions.md
jbicker: "You can still manage your account" it is.

PR #440, opened 2020-01-27
Minor tweaks to python-in-gitpod

JesterOrNot:
So minor tweaks to python-in-gitpod including a reference to help with Python GUI applications

### DIFF
--- a/src/docs/python-in-gitpod.md
+++ b/src/docs/python-in-gitpod.md
@@ -2,11 +2,11 @@
 
 Gitpod comes with great support for Python builtin. Still, depending on your project, you might want to further optimize the experience.
 
-## Examples
+## Python Project Examples
 
 Before we get started, here are some examples of already-[gitpodified](https://www.gitpod.io/blog/gitpodify/) repositories!
 
-| Name                                                                                                |                                                                                                              Description |                                        Try it |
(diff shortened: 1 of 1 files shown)


src/docs/python-in-gitpod.md
```diff
@@ -137,4 +137,5 @@ To see a basic repository with Python debugging enabled, please check out [Jeste
 
 
 ## Resources
-* ***[VSCode documentation for Python debugging](https://code.visualstudio.com/docs/python/debugging)*** All the information here should apply to Gitpod as well
+* ***[VSCode documentation for Python debugging](https://code.visualstudio.com/docs/python/debugging)*** All the information here should also apply to Gitpod as well 
+* ***[Troubleshooting-Python-GUI](https://github.com/gitpod-io/gitpod/issues/795)*** Here is how to troubleshoot some vnc issues for Python GUI applications
```
jankeromnes: Interesting. I think it's great to link to relevant GitHub issues and/or Discourse posts.

However, I don't think this issue really helps me if I have "some VNC issues", right? (It's only helpful if I want to run Matplotlib with TclTk no?)

(Also, when linking to issues, I think it's nice to link to one particularly useful comment if possible, instead of linking to the entire thread, because that's a lot to read for new users.)

PR #457, opened 2020-01-31
[nav] Make "Log In" button open Gitpod login page, not workspaces

jankeromnes:
Currently, you need to click on a "Log In" button twice, which is annoying / surprising.


jankeromnes:
@JesterOrNot Please take a look when you have time :)

jankeromnes:
Awesome, thanks! 🙏

PR #464, opened 2020-02-05
rustlang: improve note about emojis in crates extension

Kreyren:


JesterOrNot:
Can we remove the `arguably annoying` part?

PR #474, opened 2020-02-07
[docs] Add python version section

JesterOrNot:


src/docs/python-in-gitpod.md
```diff
@@ -17,6 +17,10 @@ Before we get started, here are some examples of already-[gitpodified](https://w
 
 </div>
 
+## Python version
```
jankeromnes:
```suggestion
## Python Versions
```

src/docs/python-in-gitpod.md
```diff
@@ -17,6 +17,10 @@ Before we get started, here are some examples of already-[gitpodified](https://w
...
+
+By default Python Versions: `3.7.6`, `2.7.16`, `2.7.17`, and `3.7.3` are installed in Gitpod, the easiest way to install a new Python version is to use `pyenv install <VERSION>` for an example suppose you wanted to install Python `3.6.4` you would run `pyenv install 3.6.4`. You can switch between Python versions using `pyenv global <VERSION>` for example if you wanted to switch to Python version `2.7.17` you would run `pyenv global 2.7.17` Also if you are using Python 3 you must use `python3` and `pip3` instead of `python` and `pip`
```
jankeromnes: Wow, this is crazy. Good catch! I don't think we need to have 4 versions of Python pre-installed. Could you please see if you can reduce that to just 2? (One `2.x` and one `3.x`.)

Also, phrasing suggestion:

```suggestion
Gitpod workspaces come with Python versions: `3.7.6`, `2.7.16`, `2.7.17`, and `3.7.3` pre-installed by default.

The easiest way to install a new Python version is to use `pyenv install <VERSION>`. For example, suppose you wanted to install Python `3.6.4`, you would run `pyenv install 3.6.4`.

You can switch between Python versions using `pyenv global <VERSION>`. For example, if you wanted to switch to Python version `2.7.17` you would run `pyenv global 2.7.17`.
```

jankeromnes: Also, I've removed the sentence about using `python3` and `pip3`, because it's not quite right. If you want to use Python 3, you can also run `pyenv local 3.7.6`, and then just use `python` and `pip` as usual (`pyenv` makes it work right).

Also, you only want `pyenv global <VERSION>` in a Dockerfile. In a workspace, I think you'll prefer `pyenv local <VERSION>`, because that will stay effective even after stopping and restarting a workspace (contrary to `global`).

JesterOrNot:
I'm going to merge this as is!

jankeromnes:
Oops, I didn't approve this Pull Request yet, as it still requires significant changes. Maybe you meant to merge another PR?

JesterOrNot:
I'm sorry!

JesterOrNot:
😨 Can we unmerge it or make a follow up PR?

src/docs/python-in-gitpod.md
```diff
@@ -17,6 +17,14 @@ Before we get started, here are some examples of already-[gitpodified](https://w
 
 </div>
 
+## Python Versions
+
+Gitpod workspaces come with Python versions: `3.7.6`, `2.7.16`, `2.7.17`, and `3.7.3` pre-installed by default.
```
jankeromnes: The version order doesn't make sense to me (i.e. it's not sorted).

Also, there should be only one `2.x.y` version, and one `3.x.y` version. So two in total. (It's a bug in `gitpod/workspace-full` if there are more than that.)

jankeromnes:
Additionally, this part of my earlier comment still applies:

> Also, you only want pyenv global <VERSION> in a Dockerfile. In a workspace, I think you'll prefer pyenv local <VERSION>, because that will stay effective even after stopping and restarting a workspace (contrary to global).

Please address these elements. Since this PR is already merged, you can address them in a follow-up PR, but please just don't forget about these comments. 🙂

jankeromnes:
> I'm sorry!
> 😨 Can we unmerge it or make a follow up PR?

It's not a big deal 😊 these things can happen. It's not too bad, so there is no need to un-merge the PR, or to revert the commit or anything.

Please just make sure that remaining items are indeed all fully addressed in a follow-up PR, e.g. some time next week (nothing urgent).

PR #475, opened 2020-02-07
[docs] Patch python version section

JesterOrNot:
See #474

### DIFF
--- a/src/docs/python-in-gitpod.md
+++ b/src/docs/python-in-gitpod.md
@@ -23,7 +23,7 @@ Gitpod workspaces come with Python versions: `2.7.17`, and `3.7.3` pre-installed
 
 The easiest way to install a new Python version is to use `pyenv install <VERSION>`. For example, suppose you wanted to install Python `3.6.4`, you would run `pyenv install 3.6.4`.
 
-You can switch between Python versions using `pyenv local <VERSION>`. For example, if you wanted to switch to Python version `2.7.17` you would run `pyenv local 2.7.17`, however when running it in a Dockerfile you will want to replace local with global.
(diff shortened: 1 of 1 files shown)


src/docs/python-in-gitpod.md
```diff
@@ -19,11 +19,11 @@ Before we get started, here are some examples of already-[gitpodified](https://w
 
 ## Python Versions
 
-Gitpod workspaces come with Python versions: `3.7.6`, `2.7.16`, `2.7.17`, and `3.7.3` pre-installed by default.
+Gitpod workspaces come with Python versions: `2.7.17`, and `3.7.3` pre-installed by default.
 
 The easiest way to install a new Python version is to use `pyenv install <VERSION>`. For example, suppose you wanted to install Python `3.6.4`, you would run `pyenv install 3.6.4`.
 
-You can switch between Python versions using `pyenv global <VERSION>`. For example, if you wanted to switch to Python version `2.7.17` you would run `pyenv global 2.7.17`.
+You can switch between Python versions using `pyenv local <VERSION>`. For example, if you wanted to switch to Python version `2.7.17` you would run `pyenv local 2.7.17`, however when running it in a Dockerfile you will want to replace local with global.
```
jankeromnes:
```suggestion
You can switch between Python versions using `pyenv local <VERSION>`. For example, if you wanted to switch to Python version `2.7.17` you would run `pyenv local 2.7.17`. This will create a `.python-version` file that controls which Python version is the default in your project. (Note: When running it in a Dockerfile, replace `local` with `global` in order to change the system default.)
```

JesterOrNot:
K, Great Thanks Merging!

PR #476, opened 2020-02-09
[docs] make languages-and-frameworks more accessible

JesterOrNot:
Make the languages and frameworks more accessible to blind people by adding alts to images so that text readers can tell the reader what the image is of!  :octocat:


jankeromnes:
Awesome improvement, many thanks! 💯

Also, @nisarhassan12 already approved. Thanks for the review! 💯

PR #479, opened 2020-02-12
[docs] Simplify .gitpod.yml tasks syntax

jankeromnes:
(Needs to be double-checked & tested.)

### DIFF
--- a/src/blog/developing-nuxtjs-in-browser.md
+++ b/src/blog/developing-nuxtjs-in-browser.md
@@ -73,8 +73,8 @@ Here’s the contents:
       - port: 3000
         protocol: "http"
     tasks:
-      - command: >
-            yarn install &&
+      - command: |
+            yarn install
             yarn dev
 
 ## Dive deeper into Nuxt.js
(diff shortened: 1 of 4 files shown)


jankeromnes:
Thanks Sven! 💯

I'm not too comfortable yet with `|` multi-line commands yet. I hope they don't have any unforeseen side-effects in Gitpod.

But if they work, they can make multi-line commands much simpler for our users (i.e. no need to worry about adding `&&` at the end of each line except the last one).

jankeromnes:
One gotcha is that commands below a failed command will still execute:

```yml
tasks:
  - init: touch hello2
    command: |
      touch hello3
      touch hello4
      mkdir /usr/bin2
      touch hello5
```

```
$ HISTFILE=/workspace/.gitpod/cmd-0 history -r; {
> touch hello2
> } && {
> touch hello3
> touch hello4
> mkdir /usr/bin2
> touch hello5
> 
> }
mkdir: cannot create directory ‘/usr/bin2’: Permission denied

$ ls
hello  hello2  hello3  hello4  hello5
```

(i.e. with the previous `&&`-based syntax, `hello5` would not exist).

But I think that's an acceptable trade-off for convenience.

jankeromnes:
One situation where this might be problematic is if you expect one command to produce some state, and it fails, but then below you have another command that outputs a ton of logs.

You may miss the initial error, and be surprised why the expected state isn't there.

PR #513, opened 2020-03-02
[docs] Add Java Debugging section

JesterOrNot:
See #383


svenefftinge:
Nice! Would be great to also add a pointer to https://code.visualstudio.com/docs/java/java-debugging at the end as a 'read more'.

PR #537, opened 2020-03-13
[docs] make languages-and-frameworks its own folder for maintainability

JesterOrNot:
This will make it easier to work with as the number of guides will only grow from here


JesterOrNot:
Done! Mind taking another look?

netlify.toml
```diff
@@ -323,3 +323,67 @@
   to = "/docs/self-hosted/latest/install/workspaces/"
   status = 301 
 
+[[redirects]]
+  from = "/docs/bash-in-gitpod"
+  to = "/docs/languages/bash"
```
jankeromnes: While the `from` URL looks correct, the `to` URL should be canonical.

You can find out what the canonical URL of a page is by visiting it, and seeing what final URL the browser takes you to (or, you can open the browser devtools, and search for the `<link rel="canonical">` tag). Usually in the Gitpod website, the canonical URLs have a final `/` at the end (this is important). Please save visitors an unnecessary HTTP roundtrip by redirecting to the final / canonical URL in one go.

TL;DR Please add a trailing `/` after every `to` URL in this file (but the `from` URLs can stay without a slash, so as to match both with/without slash in "bad" requests).

src/docs/languages/languages-and-frameworks.md
```diff
@@ -0,0 +1,17 @@
+# Languages & Frameworks
+
+Below is a list of language and framework specific tips & tricks.
+
+- [Python](/docs/languages/python/)
+- [HTML](/docs/languages/html/)
```
jankeromnes: The menu says this is called `HTML/CSS`, not just `HTML`:

```suggestion
- [HTML/CSS](/docs/languages/html/)
```

src/docs/menu.ts
```diff
@@ -91,59 +91,59 @@ export const MENU: MenuEntry[] = [
...
-                "ruby-in-gitpod"
+                "languages/ruby"
             ),
             M(
                 "PHP",
-                "php-in-gitpod"
+                "languages/php"
             ),
             M(
                 "Vue.js",
```
jankeromnes: Side note: I've always wondered why we call this page "Vue.js" and not just "Vue".

- "Vue.js" seems to be the JavaScript framework
- "Vue" seems to be the common name of the framework, and/or a dedicated language used for Vue.js templates

src/docs/menu.ts
```diff
@@ -91,59 +91,59 @@ export const MENU: MenuEntry[] = [
     ),
     M(
         "Languages & Frameworks",
-        "languages-and-frameworks",
+        "languages/languages-and-frameworks",
```
jankeromnes: Personal preference: Please don't move the `languages-and-frameworks.md` page under the new sub-folder.

(Maybe one day we'll also have a `frameworks` sub-folder.)

src/docs/languages/languages-and-frameworks.md
```diff
@@ -0,0 +1,17 @@
...
+- [HTML/CSS](/docs/languages/html/)
+- [Java](/docs/languages/java/)
+- [C++](/docs/languages/cpp/)
+- [Go](/docs/languages/go/)
+- [Bash](/docs/languages/bash/)
+- [Ruby](/docs/languages/ruby)
+- [PHP](/docs/languages/php)
+- [Vue](/docs/languages/vue)
+- [Rust](/docs/languages/rust/)
+- [.NET](/docs/languages/dotnet)
```
jankeromnes: A few missing final `/`s:

```suggestion
- [Ruby](/docs/languages/ruby/)
- [PHP](/docs/languages/php/)
- [Vue](/docs/languages/vue/)
- [Rust](/docs/languages/rust/)
- [.NET](/docs/languages/dotnet/)
```

src/docs/menu.ts
JesterOrNot: Done!

PR #550, opened 2020-03-18
Refactorings to make the code more maintainable.

nisarhassan12:


src/components/pricing/Bussiness.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
```
JesterOrNot: The name of this file is spelled incorrectly, Buissiness -> Business

src/components/features/StartForFree.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
```
JesterOrNot: Small nit-pick, maybe a simpler name for the component? Possibly StartFree

src/components/index/Banner.tsx
```diff
@@ -0,0 +1,160 @@
+import React from 'react'
+
```
JesterOrNot: There seems to be 2 Banner.tsx files one here and one in src/components, why is this?

nisarhassan12: Thanks! this one is specific to index page and the other one is used across various pages.

JesterOrNot: Got it

PR #553, opened 2020-03-19
[docs] Refine PHP Debugging section

JesterOrNot:
This PR makes a few changes, First, for some reason, I had a VSCode specific video. So I replaced it with a Gitpod specific one. Second I removed the unneeded nesting with the VSCode extensions header which since there is only one extension -- the one that provides debugging support I replaced it with Debugging PHP in Gitpod. Finally, I added the correct configuration as a snippet which I forgot to do :sweat_smile:.


JesterOrNot:
Thanks, merging!

PR #555, opened 2020-03-24
[docs] Updated configuration intro

svenefftinge:


src/docs/configuration.md
```diff
@@ -1,12 +1,66 @@
...
+
+### Additional Information
 
-  * [.gitpod.yml](/docs/config-gitpod-file/)
-  * [Docker Image](/docs/config-docker/)
   * [Exposing Ports](/docs/config-ports/)
-  * [Start Script](/docs/config-start-tasks/)
   * [VS Code Extensions](/docs/vscode-extensions/)
-  * [Workspace Location](/docs/checkout-location/)
-  * [Prebuilt Workspaces](/docs/prebuilds/)
```
jankeromnes: I would keep most of these links, even if they're already listed above. Just in case you didn't click while reading through.

jankeromnes:
On a side-note, it's a bit unclear if we address project maintainers or occasional contributors here, but I guess we're addressing both so that's expected.

PR #565, opened 2020-04-01
Minor fixes to start tasks documentation

JesterOrNot:
Minor fixes to start tasks documentation

- Syntax errors: both the openMode table and the openIn table had no space between the colon and the value i.e. `openMode:split-left` causes a syntax error so I replaced it with `openMode: split-left`
- Outdated section: early in the docs we say here is our setup but we have a second part of the init section, so I added it
- Minor fixes: in the before task section we have the following
```yaml
tasks:
  - before: setup.sh
    init: npm install
    command: npm run dev
```
I sincerely doubt that `setup.sh` is in `$PATH` so I changed it to `./setup.sh`

### DIFF
--- a/src/docs/config-start-tasks.md
+++ b/src/docs/config-start-tasks.md
@@ -10,8 +10,10 @@ For instance, the start script for the Gitpod documentation repository is define
 
 ```yaml
 tasks:
-- init: yarn && yarn build
-  command: yarn dev --host 0.0.0.0
+  - init: yarn && yarn build
+    command: yarn dev --host 0.0.0.0
+  - init: cd plugins/gatsby-remark-gitpod && npm install
+    openMode: split-right
 ```
 
 You can have multiple tasks, which are opened on separated terminals.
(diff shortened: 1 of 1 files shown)


src/docs/config-start-tasks.md
```diff
@@ -10,8 +10,10 @@ For instance, the start script for the Gitpod documentation repository is define
 
 ```yaml
 tasks:
-- init: yarn && yarn build
-  command: yarn dev --host 0.0.0.0
+  - init: yarn && yarn build
+    command: yarn dev --host 0.0.0.0
+  - init: cd plugins/gatsby-remark-gitpod && npm install
+    openMode: split-right
```
jankeromnes, line 16: I never understood why in the Gitpod website repo we open a split Terminal that just does `npm install` in a weird Gatsby plugin directory.

To me it doesn't seem important enough that it should get so much screen real-estate (a full visible half-terminal).

(Also, in general I don't like split terminals, because each terminal becomes too narrow for my taste, but it's probably just me.)

JesterOrNot: Fixed

JesterOrNot:
Thanks, Merging 🚀

PR #569, opened 2020-04-02
Change openMode for plugin terminal to tab-after

JesterOrNot:
This is in response to https://github.com/gitpod-io/website/pull/565#discussion_r401675905

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,7 +6,6 @@ tasks:
   - init: yarn && yarn build
     command: yarn dev --host 0.0.0.0
   - init: cd plugins/gatsby-remark-gitpod && npm install
-    openMode: split-right
 
 vscode:
   extensions:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -3,9 +3,9 @@ ports:
     onOpen: open-preview
 
 tasks:
+  - init: cd plugins/gatsby-remark-gitpod && npm install
   - init: yarn && yarn build
     command: yarn dev --host 0.0.0.0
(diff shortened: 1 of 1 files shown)


.gitpod.yml
```diff
@@ -6,7 +6,7 @@ tasks:
   - init: yarn && yarn build
     command: yarn dev --host 0.0.0.0
   - init: cd plugins/gatsby-remark-gitpod && npm install
-    openMode: split-right
+    openMode: tab-after
```
jankeromnes, line 9: Isn't `tab-after` the default behavior? If so, please remove this line. 👍

Or, if your goal is to make this not be the default visible terminal, please move it above the other terminal, which will then itself become the default visible terminal.

JesterOrNot: Done!

jankeromnes:
@nisarhassan12 I'm curious, how do you feel about this suggested change? 🙂

.gitpod.yml
```diff
@@ -6,7 +6,6 @@ tasks:
   - init: yarn && yarn build
     command: yarn dev --host 0.0.0.0
   - init: cd plugins/gatsby-remark-gitpod && npm install
```
nisarhassan12, line 8: Thanks! LGTM :+1: Maybe re-order this so that the terminal that runs `yarn dev` is not hidden?
```suggestion
  - init: cd plugins/gatsby-remark-gitpod && npm install
  - init: yarn && yarn build
    command: yarn dev --host 0.0.0.0
```

PR #570, opened 2020-04-02
Add logos of more organizations that trust Gitpod

jankeromnes:
ToDo:
- [x] Finalize the list (add / remove / re-order logos)
- [x] Make logos grey

Note: this is a temporary quick-fix, because @nisarhassan12 is already working on a better front-page design. But I like this list of logos -- maybe they can be an inspiration for the upcoming new design.


jankeromnes:
Currently looks like this (WIP):

Screenshot 2020-04-02 at 16 41 31

jankeromnes:
Ok, this looks good now. Merging as is (without review) because this is temporary anyway (new design is coming next week).

But, meanwhile, I'm happy to make any adjustments if you have feedback.

Issue #575, opened 2020-04-06
Adjust pricing 


ChristinFrohne:
Problem:

- team subscription isn't clear
- missing option to buy a subscription immediately
- free trial is poorly communicated
- add feature of booking extra hours to personal plan (let's leave the tick out of the feature list until it's actually enabled) 

<img width="963" alt="Bildschirmfoto 2020-04-06 um 11 36 23" src="https://user-images.githubusercontent.com/46345125/78545072-71534280-77fb-11ea-986d-af1f76885be6.png">

Could you look into it when you find time @nisarhassan12 ?

nisarhassan12:
Sure! I'm currently working on #560 there are just a few things remaining I will have a look at it afterwards please let me know it its urgent.

ChristinFrohne:
It's not urgent and fine if you have a look at it after the current issue. Thanks!

nisarhassan12:
I have added **Opt. Extra Hours" in the features list 

![image](https://user-images.githubusercontent.com/46004116/78626797-4293a580-78a9-11ea-9cf1-58b5268bd542.png)

What should I put in the description that appears once the question mark is hovered?

nisarhassan12:
Also where this Create Team link is gonna point to 
![image](https://user-images.githubusercontent.com/46004116/78628071-994eae80-78ac-11ea-9358-8655ec1c3ab4.png)

ChristinFrohne:
>I have added "Opt. Extra Hours" in the features list
>(let's leave the tick out of the feature list until it's actually enabled)

Extra hours have not been enabled yet. Let's add it at a later point to the list.

>Also where this Create Team link is gonna point to

https://gitpod.io/teams/

PR #590, opened 2020-04-21
[pricing] small adjustments.

nisarhassan12:
Fixes gitpod-io/website#575

This is how it looks:

image

The same pricing box component is used in other places as well so the changes that I made are not scoped to just pricing.

image

### DIFF
--- a/src/components/PopOver.tsx
+++ b/src/components/PopOver.tsx
@@ -34,24 +34,37 @@ const StyledDescripion = styled.span`
         background: ${colors.offWhite2};
         z-index: 1000 !important;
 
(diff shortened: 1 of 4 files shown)


ChristinFrohne:
Thanks for looking into it. 

- [ ] Could you delete "Team Manageable" for the Professional and Unlimited Plan and move the tooltip to `create team`? I don't think it makes sense to mention the team option twice

- [ ] then you should have enough space to reduce the height of the boxes a bit

nisarhassan12:
@ChristinFrohne I have made the changes, please have a look again.

Issue #600, opened 2020-04-29
enabling horizontal scrolling for logos


ChristinFrohne:
@jankeromnes suggested to use horizontal scrolling instead of clicking through the "trusted by" logos. Could you implement that @nisarhassan12? 

Also, the spacing between the logos is inconsistent. Could you adjust that?

<img width="1119" alt="Bildschirmfoto 2020-04-29 um 14 16 57" src="https://user-images.githubusercontent.com/46345125/80594885-1e4c6400-8a24-11ea-8247-59ce95ea4267.png">

nisarhassan12:
No longer relevant to the new website design.

PR #602, opened 2020-04-29
[TrustedBy] enable horizontal scrolling.

nisarhassan12:
This is how it looks:

image

Fixes gitpod-io/website#600


ChristinFrohne:
The scrolling bar that you have on your screenshot and the one I get on my mac look nice. But on my windows it doesn't look that good. It quite massive. Can that be adjusted? 

Annotation 2020-04-30 082839

Also could you even out the spacing between the logos? There is a lot more space before "openAI" than before "d-wave". Maybe then it's also possible to not have the "openai" logo cut

nisarhassan12:
> Also could you even out the spacing between the logos? There is a lot more space before "openAI" than before "d-wave". Maybe then it's also possible to not have the "openai" logo cut

It's quite hard to do as they are not equal in width and a few of them have whitespace in them on either side see the `Uber` logo below: 

image

I have reduced there width this is how they look now.

image

ChristinFrohne:
The spacing is better now, it's great that the logo isn't cut anymore. But I agree with Sven that the scrolling bar looks weird, especially on windows. Do you have another idea to visualize the scrolling @nisarhassan12 ? Otherwise, we can just stick with the arrow in the way we currently have it

svenefftinge:
Or two rows and make the logos a tad smaller?

ChristinFrohne:
I find two rows of logos on the start screen rather noisy. Another solution could be this:
Bildschirmfoto 2020-05-06 um 12 24 07

It swipes through the logos automatically and you can also click through it via the dots. Here we could arrange 3 sets of 5 logos each and swipe through them.
https://buddy.works/on-premises

nisarhassan12:
> I find two rows of logos on the start screen rather noisy. Another solution could be this:
> Bildschirmfoto 2020-05-06 um 12 24 07
> 
> It swipes through the logos automatically and you can also click through it via the dots. Here we could arrange 3 sets of 5 logos each and swipe through them.
> https://buddy.works/on-premises

Thanks @ChristinFrohne It looks nice but it can be problematic to use on mobile devices i.e the tap targets are gonna be too small.

ChristinFrohne:
Have you checked their mobile view? They use it anyway.
Maybe we could enable finger swiping on mobile instead of clicking on the dots.

nisarhassan12:
Thanks.Swiping would make sense. thanks.

PR #606, opened 2020-04-30
[self-hosted] make all links visible in menu

### LINKED ISSUE #604: [self-hosted] make all pages reachable via menu

JesterOrNot:
Fixes #604


meysholdt:
Thank you!

let's not yet show the following menu items:
* "admin": the contents is not yet ready
* "helm": it's not about Gitpod, but just in case somebody has trouble with helm.

PR #616, opened 2020-05-06
[self-hosted] More messaging tweaks

jankeromnes:


src/pages/self-hosted.tsx
```diff
@@ -46,9 +46,9 @@ const SelfHostedPage: React.SFC<{}> = () => (
                 <Banner
                     subtitle="Self-hosted Gitpod"
                     title={<h1>
-                        Keep your data on your network,
+                        Keep Your Data on Your Network,
                             <br />
-                        Host Gitpod on <strong>your own infrastructure</strong>.
+                        Host Gitpod on <strong>Your Own Infrastructure</strong>.
```
ChristinFrohne, line 51: I see what you mean, the headline got very long. It'd be nice if it fits in 2 lines instead of three. What if we make the font size smaller? Using the size of h2?

jankeromnes: Or maybe we use `Infra` instead of `Infrastructure`? That seems to be common.

ChristinFrohne: Or use another word? `server`?

jankeromnes: Then it would be `Servers` plural. Or `on Your Own Cluster`. But I think the most accurate word is still `Infrastructure`.

jankeromnes:
Rebased on top of #617.

jankeromnes:
I'm happy with the result:
- https://deploy-preview-616--gitpod-website.netlify.app/
- https://deploy-preview-616--gitpod-website.netlify.app/pricing/
- https://deploy-preview-616--gitpod-website.netlify.app/self-hosted/

src/components/Nav.tsx
```diff
@@ -268,8 +268,8 @@ class Nav extends React.Component {
                         
                             <ul className={`nav__items ${isNavRendered ? 'navIsRendered' : 'navIsNotRendered'}`} >
                                 <li className="nav__item"><Link activeClassName="active" to='/features/' className="link">Features</Link></li>
-                                <li className="nav__item"><Link activeClassName="active" to='/self-hosted/' className="link">Self-Hosted</Link></li>
```
svenefftinge, line 271: I changed the order on purpose. why do you change it back?

src/components/self-hosted/SelfHostedPlans.tsx
```diff
@@ -48,16 +48,16 @@ const plans: SelfHostedPlanProps[] = [
     {
         title: 'Free',
         subtitle: 'For up to 5 users',
```
svenefftinge, line 50: this file is no longer used anywhere. should be deleted

src/components/Nav.tsx
jankeromnes: Because it doesn't feel quite right:

- Features (of gitpod.io mostly)
- Self-Hosted
- Pricing (of gitpod.io and self-hosted)

Pricing is one of our top-visited pages, i.e. a lot of people find it interesting. Pushing something in front of it probably doesn't increase visits much, but feels like we're trying hard to push users to use Self-Hosted.

I'm happy to restore your order if you feel strongly about it though. 🙂

svenefftinge: Yes, please restore. We want to emphasize the 'self-hosted' thing. So we first we what, and then we talk about pricing. Also see e.g. https://bitbucket.org/product

jankeromnes: Restoring

svenefftinge:
> Bildschirmfoto 2020-05-06 um 14 40 14
> 
> `install now`should link to docs/self-hosted, shouldn't it?
> Looks great. Thanks!

no, it should link to the visually much clearer section below.

jankeromnes:
All good now! 🎉 Many thanks. 👍

PR #636, opened 2020-05-16
lock the overflow-y when the nav items are shown on mobile.

nisarhassan12:
This is how it looks:

image

Fixes gitpod-io/website#635


nisarhassan12:
Many! Thanks I agree with you I have opened an issue for this https://github.com/gitpod-io/website/issues/673

PR #637, opened 2020-05-16
[docs] Fix table scaling for languages and frameworks

### LINKED ISSUE #495: Languages and Frameworks Scaling on Mobile
As of now on mobile to get to the try it or a lot of the descriptions in the examples section
image
Can we do something with the scaling so the user can have a general overview without any un-needed interaction?

JesterOrNot:
Fixes #495


src/docs/languages/bash.md
```diff
@@ -6,7 +6,7 @@ Bash is the default shell for Gitpod. However, if you're developing a Bash scrip
 
 Here are a few Bash example projects that are already automated with Gitpod:
 
-<div class="table-container">
+<div class="language-examples">
```
nisarhassan12, line 9: Please re-add the `table-container` class that was in charge making the table look good on small devices.
```suggestion
<div class="table-container language-examples">
```
With its removal, this is how the table looks on small devices:

image

the buttons can no longer be accessible on mobile.

src/docs/languages/cpp.md
```diff
@@ -6,7 +6,7 @@ Gitpod supports C++ right out of the box, but there are still ways to enhance yo
 
 Here are a few C++ example projects that are already automated with Gitpod:
 
-<div class="table-container">
+<div class="language-examples">
```
nisarhassan12, line 9: Please re-add the close as told above 
```suggestion
<div class="table-container language-examples">
```

src/docs/languages/html.md
```diff
@@ -6,10 +6,15 @@
 
 Here are a few HTML/CSS example projects that are already automated with Gitpod:
 
-Repository | Description | Try it
----|---|---
-[Front-End-Checklist](https://github.com/thedaviddias/Front-End-Checklist) | The perfect Front-End Checklist for modern websites and meticulous developers | [![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/thedaviddias/Front-End-Checklist)
-[Devhints](https://github.com/rstacruz/cheatsheets) | TL;DR for developer documentation - a ridiculous collection of cheatsheets | [![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/rstacruz/cheatsheets)
+
+<div class="language-examples">
```
nisarhassan12, line 10: I didn't know that we were missing the `div` with the `table-container` class here.

src/docs/languages/go.md
```diff
@@ -6,7 +6,7 @@ Gitpod supports Go right out of the box, but there are still ways to optimize yo
 
 Here are a few Go example projects that are already automated with Gitpod:
 
-<div class="table-container">
+<div class="language-examples">
```
nisarhassan12, line 9: Please, add the `table-container` class in all the places from where it got removed.

src/styles/normalize.ts
```diff
@@ -436,6 +436,11 @@ export default `
         }
     }
 
+    .language-examples td,
+    .language-examples tr {
+        font-size: 0.915em;
```
nisarhassan12, line 441:
```suggestion
        font-size: 0.915rem;
```

nisarhassan12:
> Currently, I don't know of a solution via which we can align the tables so that users don't have to use the scrollbars to view the offscreen contents of the table.

@ChristinFrohne Do you have any ideas on How way may achieve this? sites usually use horizontal scroll for tables when the device width is small.

ChristinFrohne:
>Currently, I don't know of a solution via which we can align the tables so that users don't have to use the scrollbars to view the offscreen contents of the table.
>@ChristinFrohne Do you have any ideas on How way may achieve this? sites usually use horizontal scroll for tables when the device width is small.

What about having only two cells on mobile and just adding the `open in Gitpod`button underneath each description?

src/docs/languages/go.md
JesterOrNot: done!

nisarhassan12: Thanks.

PR #645, opened 2020-05-18
firebase: init

Kreyren:


src/docs/frameworks/firebase.md
```diff
@@ -0,0 +1,13 @@
+![](https://raw.githubusercontent.com/gitpod-io/firebase-on-gitpod/master/img/firebase-on-gitpod-header.png)
```
jankeromnes, line 1: Please add an alt image description, e.g.:

```suggestion
![Gitpod on Firebase, a proof of concept](https://raw.githubusercontent.com/gitpod-io/firebase-on-gitpod/master/img/firebase-on-gitpod-header.png)
```

However, since none of the other guides have a header, maybe having one here isn't necessary.

src/docs/frameworks/firebase.md
```diff
@@ -0,0 +1,13 @@
...
+
+# Firebase on Gitpod
+
+Firebase is a platform owned by Google which is used to develop mobile and web applications
+
+[![Open proof-of-concept in gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io#snapshot/45abf388-128e-4cce-b336-20ba69288d84)
```
jankeromnes, line 7: Please explain what this button does, e.g. by adding an intro sentence like:

```suggestion
To experience Firebase in Gitpod, we have created a proof-of-concept snapshot that you can launch in one click:

[![Open Firebase proof-of-concept in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io#snapshot/45abf388-128e-4cce-b336-20ba69288d84)
```

src/docs/frameworks/firebase.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Configuring the access token
```
jankeromnes, line 9: Before this section, I think it would be cool to explain how one can install Firebase in Gitpod.

E.g. assume that the reader is relatively new to Gitpod -- the Scala section is a great example of how to achieve this: https://www.gitpod.io/docs/languages/scala/#installing-scala

src/docs/frameworks/firebase.md
```diff
@@ -0,0 +1,13 @@
...
+
+![Theia (Gitpod's editor)](https://github.com/eclipse-theia/theia) is as of 18/05/2020 using `iframe` to display pages in it's preview which is not allowed on some Google websites (https://github.com/gitpod-io/gitpod/issues/1528) and so firebase on gitpod requires the end-user to be logged-in using a hyperlink generated from `firebase login --no-localhost` and it's token stored as environmental variable `FIREBASE_TOKEN` in https://gitpod.io/settings to avoid parsing `firebase subcommand --token YOUR_TOKEN` for every command.
```
jankeromnes, line 11: Theia is not really an editor, but a full IDE:

```suggestion
![Theia (Gitpod's IDE)](https://github.com/eclipse-theia/theia) is as of 18/05/2020 using `iframe` to display pages in its preview which is not allowed on some Google websites (https://github.com/gitpod-io/gitpod/issues/1528) and so Firebase on Gitpod requires the end-user to log in using a hyperlink generated from `firebase login --no-localhost` and its token stored as environmental variable `FIREBASE_TOKEN` in https://gitpod.io/settings/ to avoid passing `--token YOUR_TOKEN` for every `firebase` command.
```

src/docs/menu.ts
```diff
@@ -145,6 +145,10 @@ export const MENU: MenuEntry[] = [
                 "Julia",
                 "languages/julia"
             ),
+            M(
+                "Firebase",
+                "frameworks/firebase"
+            ),
```
jankeromnes, line 151: This contribution raises interesting questions about the "Languages & Frameworks" section.

- Initially, the section was meant to indeed provide how-to guides for common languages and frameworks.
- When we had covered most languages, and started adding frameworks, it was confusing to mix frameworks with languages (e.g. a list like "C++, Ruby, Java, Pandas" doesn't make sense, because you're mixing languages and non-languages). So we decided to merge the frameworks into their specific language page (i.e. merge "Pandas" into the "Python" guide)
- However, the "Firebase" framework is not strongly related to one particular language, is it?

I still find it confusing to mix frameworks with languages, but we also can't merge Firebase into an existing language guide.

Maybe we should indeed have separate "Languages" and "Frameworks" sub-menus? Also, Firebase isn't even a "framework". Maybe it's more like a "tool", or "technology"? (Same goes for databases and such). So maybe let's have "Languages" and "Tools" sub-menus?

PR #650, opened 2020-05-19
Update images in static folder.

JesterOrNot:
On opening workspaces, these appear to be modified


jankeromnes:
I think they're automatically copied over from https://github.com/gitpod-io/website/tree/master/src/docs/release-notes

PR #652, opened 2020-05-20
remove the newsletter form for now as the functionality does not work.

nisarhassan12:


nisarhassan12:
Thanks! @JesterOrNot I think this is worth fixing but would make sense to have to remove from the production site unless it's fixed.

JesterOrNot:
In that case, let's add a maintenance message saying something like hey 'the newsletter is under maintenance please check back soon!'

JesterOrNot:
And let's comment it out instead of flat out removing it.

nisarhassan12:
Thanks! commenting is a good idea. 

> In that case, let's add a maintenance message saying something like hey 'the newsletter is under maintenance please check back soon!

Let's open a Github issue for this instead as discussed on Slack.

PR #668, opened 2020-05-28
docs: Use root instead of sudo when interacting with apt

Kreyren:
Relevant: https://github.com/gitpod-io/gitpod/issues/1561#issuecomment-635507762

Signed-off-by: Jacob Hrbek <kreyren@member.fsf.org>

### DIFF
--- a/src/blog/native-ui-with-vnc.md
+++ b/src/blog/native-ui-with-vnc.md
@@ -18,9 +18,11 @@ A few days ago [Gero introduced Docker builds to Gitpod](/blog/docker-in-gitpod)
 Let’s use Gitpod to write [a simple desktop-native application in Go](https://github.com/32leaves/gitpod-hello-ui-demo) using [libui](https://github.com/andlabs/libui). First, we create a GitHub repo and a [little bit of configuration](https://github.com/32leaves/gitpod-hello-ui-demo/commit/fea580735c09fa704531a810e8ec7ca6a5c03a88): we need a Dockerfile to install libui’s dependencies (see below), and we need to tell Gitpod to use that Dockerfile.
 
     FROM gitpod/workspace-full-vnc
-    RUN sudo apt-get update && \
-        sudo apt-get install -y libgtk-3-dev && \
-        sudo rm -rf /var/lib/apt/lists/*
+    RUN true \
+        && apt-get update -q && \
+        && apt-get install -qy \
+            libgtk-3-dev \
+        && rm -rf /var/lib/apt/lists/*
 
 This Dockerfile uses the gitpod/workspace-full-vnc image which sports the setup necessary to run X11 and VNC in Gitpod. It runs a [web-based VNC viewer](https://novnc.com) on port 6080 — during startup, Gitpod will ask you if you want to open this page.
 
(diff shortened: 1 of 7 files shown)


src/blog/native-ui-with-vnc.md
```diff
@@ -18,9 +18,11 @@ A few days ago [Gero introduced Docker builds to Gitpod](/blog/docker-in-gitpod)
 Let’s use Gitpod to write [a simple desktop-native application in Go](https://github.com/32leaves/gitpod-hello-ui-demo) using [libui](https://github.com/andlabs/libui). First, we create a GitHub repo and a [little bit of configuration](https://github.com/32leaves/gitpod-hello-ui-demo/commit/fea580735c09fa704531a810e8ec7ca6a5c03a88): we need a Dockerfile to install libui’s dependencies (see below), and we need to tell Gitpod to use that Dockerfile.
 
     FROM gitpod/workspace-full-vnc
-    RUN sudo apt-get update && \
-        sudo apt-get install -y libgtk-3-dev && \
-        sudo rm -rf /var/lib/apt/lists/*
+    RUN true \
```
jankeromnes, line 21: Also, what is the purpose of prefixing commands with `true &&`?

Kreyren: `RUN \` doesn't work so and `RUN : \` looks ugly.

Using that to make the code more readable

Kreyren: Also it can be used to manage caching by adding arguments to true alike

`RUN true "something" \`

Kreyren:
> Telling users to use USER root is dangerous, because in the past this has lead to users creating files and directories that belong to root:root, and thus they lost access to these in workspaces and got very confused about how to restore access.

Would agree for commands that do not handle apt, but as shown using sudo with apt on gitpod seems to result in breakage in some scenarios.

Might be a bug in layers though

Kreyren:
@jankeromnes and here is your proof-of-concept! 

https://github.com/gitpod-io/gitpod/issues/1565

thanks @ronys for reporting the issue that saved me a time since i don't have to make PoC for this merge request ^-^

corneliusludmann:
> Would agree for commands that do not handle apt, but as shown using sudo with apt on gitpod seems to result in breakage in some scenarios.

Does it? Do you have any examples?



> @jankeromnes and here is your proof-of-concept!
> 
> [gitpod-io/gitpod#1565](https://github.com/gitpod-io/gitpod/issues/1565)

This has nothing to do with `sudo`. The warning
```
sudo: setrlimit(RLIMIT_CORE): Operation not permitted
```
can be ignored safely. See https://github.com/sudo-project/sudo/issues/42.

I'm all for keeping `sudo` in the docs as long as there are no good reasons not to.

Kreyren:
> can be ignored safely. 

It can't because it seems to make gitpod to fail during the docker build, see https://github.com/Kreyren/gitpod-sudo-PoC

EDIT: Nevermind passes the build, but the error message is still quite confusing.. Can't we apply https://github.com/sudo-project/sudo/issues/42#issuecomment-599142606 ?

jankeromnes:
> EDIT: Nevermind passes the build, but the error message is still quite confusing.. Can't we apply https://github.com/sudo-project/sudo/issues/42#issuecomment-599142606 ?

Hey, great find @Kreyren, thanks! We should indeed probably upgrade `sudo` in Gitpod's workspace images.

Other than that, I'll close this PR, because we don't want to use `USER root` instead of `sudo`, because it encourages users to create files owned by `root` that will no longer be accessible to them in running workspaces.

PR #688, opened 2020-06-24
Fix ugly line break in "Code Reviews"

i-ky:
Newline character inside single quotes seems to be rendered verbatim, this produces an ugly line break in a very inappropriate place.


i-ky:
One may also argue that this is an issue of website generator in use, because the page looks fine on GitHub. While GitHub's markdown renderer uses the same `<code>` HTML tag for single-quoted text, it seems to replace newline characters with spaces automatically.

svenefftinge:
Thank you!

PR #694, opened 2020-07-07
pricing: add "/user/month" to cloud pricing plans.

nisarhassan12:
This is how it looks:

image

### DIFF
--- a/src/components/PopOver.tsx
+++ b/src/components/PopOver.tsx
@@ -28,38 +28,27 @@ const StyledDescripion = styled.span`
     }
 
     .description__text {
+        position: absolute;
         padding: .2rem .5rem;
         font-size: 1.2rem;
         color: ${colors.textDark};
         background: ${colors.offWhite2};
-        z-index: 1000 !important;
-
-        @media(max-width: 1080px) {
-            position: absolute;     
-            right: -5rem;
-            top: 2.4rem;
-            min-width: 18rem;
-            z-index: 1999999999;
-        }
-
-         @media(min-width: 1081px) {
-            position: absolute;     
-            top: 0;
-            left: 3rem;
-            min-width: 19rem;
-        }
-    }
-
-    .bottom {
-        font-weight: 400;
-        right: -10rem;
+        left: 4rem;
         top: 2.4rem;
         min-width: 18rem;
         z-index: 9999999999;
         transform: translateX(-16.5rem);
+    }
 
+    .bottom {
+        font-weight: 400;
+        top: 2.4rem;
+        min-width: 18rem;
+        z-index: 9999999999;
+        transform: translateX(-16.5rem);
         @media(max-width: 1081px) {
-            transform: translateX(-7.5rem);
+            transform: translateX(-4.5rem);
+            left: -8rem;
         }
     }
 `
--- a/src/components/PricingBox.tsx
+++ b/src/components/PricingBox.tsx
@@ -99,7 +99,7 @@ const StyledPricingBox = styled.div<StyledPricingBoxProps>`
         font-weight: 600;
 
         span {
-            font-size: 80%;
+            font-size: 70%;
             font-weight: 400;
         }
     }
@@ -231,6 +231,10 @@ const StyledPricingBox = styled.div<StyledPricingBoxProps>`
     .spacer {
         height: 2.5rem;
     }
+
+    .user-month {
+        font-size: 95%;
+    }
 `
 
 export interface PricingBoxProps {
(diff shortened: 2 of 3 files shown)


ChristinFrohne:
>should be on the same line as the price (i.e. $9 / user / month, not $9 and somewhere below / user / month)

for the layout it's nicer to have clear hierarchies and not to squeeze all information in one line. GitLab has a similar layout: https://about.gitlab.com/pricing/
 
>Instead of the Free price, it should say $0, because then it's much more obvious that it's free (it makes it much easier to visually compare prices across plans, because they all have the same format, and also the plan name itself is already Free, so there is no need to repeat it)

I totally agree. The name of the plan is already "free" so there is no need to repeat that.

As mentioned on Slack, I'm suggesting this layout
Bildschirmfoto 2020-07-17 um 09 25 27

jankeromnes:
> As mentioned on Slack, I'm suggesting this layout

This proposed layout would work, but as we've previously seen, any isolated price (i.e. without the word "month" on the same line) has the risk of being misinterpreted by users: We've had many questions like "is this a one-time price or is it per month?" (even if below we write in small that it's per month).

But, since GitLab presents it similarly, I'm okay with going for this design if another design is not practical right now. But I still believe that the price, and the words "user" and "month", should be on the same line (even if price is very big and bold, and "per user per month" is very small).

Also, I personally find `per user/month` slightly confusing, because `/` and `per` mean the same thing (so why are they different?) I.e. `per user per month`, or `/ user / month` is consistent and relatively easy to understand, while `per user/month` is confusing (e.g. one thought is, "what sort of unit is a `user/month`?"). But maybe that's just me. 😊

ChristinFrohne:
Sorry for the late reply

I haven't seen that spaces are being using before an dafter `/`. I checked again with other pages and Grammarly: it's without spaces
"Many people wonder whether to use a space before and after slashes in a sentence. The answer is typically no on both fronts. " [Grammerly](https://www.grammarly.com/blog/slash/#:~:text=Space%20Before%20and%20After%20Slash&text=Spaces%20before%20slashes%20should%20be,terms%20for%20ease%20of%20reading.)

>Also, I personally find per user/month slightly confusing, because / and per mean the same thing (so why are they different?

I think it's helpful to use `per user/month`:
-  `/`could also mean "and/or". By using `per`the meaning is immediately clear. Otherwise, the user could think there is an option to either pay per user or per month? 
- other platforms also use it this wording, see https://github.com/pricing
- without spaces it looks quite pressed together. By using `per`we work around the spacing issue

ChristinFrohne:
@nisarhassan12 I noticed three issues with the pricing box

1. Please leave some space between the feature and the button
Bildschirmfoto 2020-07-29 um 10 03 17

2. on tablet view the tool tip box is underneath the button. Please arrange it to the front
Bildschirmfoto 2020-07-29 um 10 03 39

3. have the same spacing between the last feature and the button on default and tablet/mobile view. Currently the spacing gets a lot bigger on smaller devices. It's okay if the tooltip partly covers the button
normal: 
Bildschirmfoto 2020-07-29 um 10 10 32

tablet:
Bildschirmfoto 2020-07-29 um 10 10 45

nisarhassan12:
@ChristinFrohne I have made the requested changes please have a look again whenever you'll have time. Thanks.

ChristinFrohne:
@nisarhassan12 could you please adjust the wording `per user/month` as suggested above. If @jankeromnes comes back from his vacation and would like to discuss more. I'm happy to do so.  

The free plan still has a lot of spacing. Could you move "create team" and the "buy now"-button a bit further up and reduce the height of the box a bit?

Thanks!

ChristinFrohne:
What are your thoughts about it @nisarhassan12 ? Would be great to complete this soon

nisarhassan12:
@ChristinFrohne This Pr is old and since it was first made we have made a significant amount of changes and there are conflicts as well so making a new Pr would make more sense. Would be nice if you can open a new issue in accordance to the current state of https://www.gitpod.io and what are the changes expected to be made ? i.e Is it just adding  `per user/month` and removing `/month` next to the price.

image

ChristinFrohne:
@nisarhassan12 the issue is still the same and the mockup hasn't changed. If you'd like to start from a new issue, could you open it yourself?

The pricing page should look like this, including "Per user/month" and semibold fonts (see image).
Please note that the previous mentioned issues of the hidden tooltip(`on tablet view the tool tip box is underneath the button.`, and the changing spacing for mobile view (`have the same spacing between the last feature and the button on default and tablet/mobile view. Currently the spacing gets a lot bigger on smaller devices.`) still exists.

Bildschirmfoto 2020-10-20 um 13 29 26

nisarhassan12:
@ChristinFrohne I have made a new Pr #815 for adding `per user/month` to the cards and for the other problems opened 2 issues which I'll probably address later on.

#814 
#813

ChristinFrohne:
@nisarhassan12 can this PR be closed?

nisarhassan12:
Yeah. Sure. I'm closing this now.

PR #700, opened 2020-07-17
delayed: use gatsby-plugin-twitter for loading the tweets.

nisarhassan12:
I have tried to find a plugin which lets us fetch tweet embeds at build time but so far. I'm unable to do this so. The only difference that using `gatsby-plugin-twitter` has made is that the markup for the tweets is now server-rendered. The loading of the tweets is still terrible there is a lot of JS that twitter is shipping.

Tweets to me look like a non-interactive element. I wonder why they are shipping too much JS.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -82,6 +82,7 @@ module.exports = {
         'gatsby-transformer-sharp',
         'gatsby-plugin-react-helmet',
         'gatsby-plugin-catch-links',
+        `gatsby-plugin-twitter`,
         {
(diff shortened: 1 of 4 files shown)


nisarhassan12:
@jankeromnes Maybe implementing their layout ourself would be a better idea i.e we can just use the tweet's markup and style it the way we like this would for sure result in a lot less JS being shipped.

nisarhassan12:
With the self-build tweets in place, this is now irrelevant.

PR #702, opened 2020-07-20
[Doc] Append brew update to brew install

corneliusludmann:
There seems to be the need to run `brew update` before running `brew install`. I guess the brew index in our workspace container becomes out of date at some time. I added `brew update` in the docu.

See also: https://community.gitpod.io/t/scala-gitpod-example-doesnt-work-invalid-formula-error/1572/2

### DIFF
--- a/src/docs/config-docker.md
+++ b/src/docs/config-docker.md
@@ -32,7 +32,7 @@ You can find the source code in <a href="https://github.com/gitpod-io/workspace-
 FROM gitpod/workspace-full
 
 # Install custom tools, runtime, etc.
-RUN brew install fzf
+RUN brew update && brew install fzf
 ```
 
 When you are launching the Gitpod IDE, the local console will use the `gitpod` user, so all local settings, config file, etc. should apply to `/home/gitpod` or be run using `USER gitpod` (we no longer recommend using `USER root`).
--- a/src/docs/languages/bash.md
+++ b/src/docs/languages/bash.md
@@ -27,7 +27,7 @@ FROM gitpod/workspace-full
 
 USER gitpod
 
-RUN brew install shellcheck
+RUN brew update && brew install shellcheck
 ```
 
 You should also install the shellcheck VS Code extension in Gitpod, by adding a [.gitpod.yml](/docs/config-gitpod-file/) configuration file to your repository that looks like this (notice the `vscode` extensions section):
--- a/src/docs/languages/go.md
+++ b/src/docs/languages/go.md
@@ -68,7 +68,7 @@ FROM gitpod/workspace-full
 
 USER gitpod
 
-RUN brew install dep
+RUN brew update && brew install dep
 ```
 
 Also, don't forget to reference the above Dockerfile in your `.gitpod.yml` configuration file, like so:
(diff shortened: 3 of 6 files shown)


jankeromnes:
> There seems to be the need to run `brew update` before running `brew install`. I guess the brew index in our workspace container becomes out of date at some time. I added `brew update` in the docu.

Note: `brew`'s default behavior is to always run `brew update` before a `brew install` automatically.

However, for some reason, `brew update` takes an impractically long time in Gitpod, so we've disabled that behavior using [`HOMEBREW_NO_AUTO_UPDATE=1`](https://github.com/gitpod-io/workspace-images/blob/7d1838ba374495e91ce0cad7a96a9ab878231d11/full/Dockerfile#L115).

So far, this enabled a reasonably fast `brew install` experience, that worked reasonably well (except on two occasions where the `brew` cache got stale and caused problems, requiring a dazzle layer cache invalidation for HomeBrew).

I think a better approach would be to implement regular invalidations for the HomeBrew Docker layer. One solution could be to auto-update some version string, similar to how we currently auto-upgrade the [Node.js](https://github.com/gitpod-io/workspace-images/commit/8c0f68d1f8410fdf6ff00fd76f54ab61125a12ea) or [Rust](https://github.com/gitpod-io/workspace-images/commit/9f714ef2f34ccf6fb116f2a35a457f008d83b7ba) layers.

corneliusludmann:
> Note: `brew`'s default behavior is to always run `brew update` before a `brew install` automatically.
> 
> However, for some reason, `brew update` takes an impractically long time in Gitpod, so we've disabled that behavior using [`HOMEBREW_NO_AUTO_UPDATE=1`](https://github.com/gitpod-io/workspace-images/blob/7d1838ba374495e91ce0cad7a96a9ab878231d11/full/Dockerfile#L115).

Oh, good to know. That this PR makes no sense. :-D

However, disabling `brew update` does not sound like a good idea to me. Even when we maintain an up-to-date brew cache for our docker containers, what If someone restarts an old workspace and would like to run `brew install`?

But, I'm not really into that topic either. Maybe you're right and we should just document the current behavior.

jankeromnes:
> However, disabling `brew update` does not sound like a good idea to me. Even when we maintain an up-to-date brew cache for our docker containers, what If someone restarts an old workspace and would like to run `brew install`?

You're right, that's a good point. Maybe a better solution would be to figure out why `brew update` is so slow on Gitpod, and make it much faster. Then we could re-enable it without problem.

corneliusludmann:
The first `brew update` on a new workspace takes ~ 3 - 4 minutes:
```
real    3m26.567s
user    0m8.481s
sys     0m17.821s
```

The second time it is much faster:
```
real    0m18.028s
user    0m0.446s
sys     0m0.879s
```

Info from the Homebrew project leader:

> > is a three-minute wait normal for the first brew up of a day? I don't remember things being anywhere near so slow in the past.
> 
> It depends on many things:
> 
> - the speed of your internet connection
> - the speed of your machine
> - the distance from GitHub
> - the load on the GitHub API
> - the load on GitHub's git layer
> - how many commits you need to fetch for each repository

https://github.com/Homebrew/brew/issues/4755#issuecomment-416572302

Issue #709, opened 2020-08-05
Blog: Layout need to be adjusted for mini Desktops and small Laptops.


nisarhassan12:
Post cards on /blog don't look that good with having a lot of space between them.

![image](https://user-images.githubusercontent.com/46004116/89400874-ee50e300-d72d-11ea-8716-772894cceefb.png)

samuelpath:
## Before with `space-between`
![image](https://user-images.githubusercontent.com/6030745/89425056-cc109280-d738-11ea-9234-a12584221c3d.png)

## After with `space-around`
![image](https://user-images.githubusercontent.com/6030745/89425103-e21e5300-d738-11ea-8535-0837bb3e9638.png)

PR #711, opened 2020-08-05
Fixes gitpod-io/website#709

samuelpath:
Improve blog layout for small width screens by using `space-around` instead of `space-between` for the blog posts container

### DIFF
--- a/src/pages/blog.tsx
+++ b/src/pages/blog.tsx
@@ -33,7 +33,7 @@ const StyledBlogPage = styled.div`
     .posts {
         display: flex;
         flex-wrap: wrap;
-        justify-content: space-between;
+        justify-content: space-around;
 
         @media (max-width: ${sizes.breakpoints.md}) {
             justify-content: space-around;
--- a/src/pages/blog.tsx
+++ b/src/pages/blog.tsx
@@ -33,7 +33,7 @@ const StyledBlogPage = styled.div`
     .posts {
         display: flex;
         flex-wrap: wrap;
-        justify-content: space-around;
+        justify-content: space-between;
 
         @media (max-width: 1140px) {
             justify-content: space-around;
--- a/src/pages/blog.tsx
+++ b/src/pages/blog.tsx
@@ -35,7 +35,7 @@ const StyledBlogPage = styled.div`
         flex-wrap: wrap;
         justify-content: space-around;
 
-        @media (max-width: ${sizes.breakpoints.md}) {
+        @media (max-width: 1140px) {
             justify-content: space-around;
         }
     }


src/pages/blog.tsx
```diff
@@ -33,7 +33,7 @@ const StyledBlogPage = styled.div`
     .posts {
         display: flex;
         flex-wrap: wrap;
-        justify-content: space-between;
+        justify-content: space-around;
```
nisarhassan12, line 36:
```suggestion
        @media(max-width: 1140px) {
            justify-content: space-around;
        }
```
Feel free to ignore this just a small nit. The Layout issue actually starts when the width of the viewport is less than or equal to  `1140px` I would recommend changing the layout only then i.e at devices with width > `1140px` the layout looks perfectly fine. I would say even better than the current change i.e before it was as follows:

image

The logo was perfectly aligned with the left of the first post.

and now with your change, it looks like this:

image

In my opinion, the alignment seems to be a bit off.

samuelpath: Thanks @nisarhassan12 for the helpful suggestion.

I see in the original code that you already had this fix in place for smaller viewports:

```
.posts {
    display: flex;
    flex-wrap: wrap;
    justify-content: space-between;

    @media (max-width: ${sizes.breakpoints.md}) {
        justify-content: space-around;
    }
}
```

However, the `space-around` started at 800px, unlike the 1140px that you suggest in your comment:

```
breakpoints: {
    sm: '400px',
    md: '800px',
    lg: '1240px',
},
```

Do you think it's good practice to add the literal value `1140px` in the `blog.tsx` file without any reference to an actual variable?

How would you do this fix by keeping the practice of using standardized variables across the codebase rather than using literal values?

Thanks.

PS. By the way, what tool do you use to show the dotted line on the webpage, as you did in your previous comment?

nisarhassan12: > However, the space-around started at 800px

Opps. Sorry I missed that.

> Do you think it's good practice to add the literal value 1140px in the blog.tsx file without any reference to an actual variable?

Thanks :slightly_smiling_face:  . I think it is ok and maybe good to do so in this case as `1140px` is used in more than 2 or 3 places.

Currently, for the site, we are using too many breakpoints and adding all of them to the variables would possibly cause bloat and it would be hard to maintain and reason about.

The reason why we are not suing just `3 - 5` breakpoints defined in the breakpoints object ( or any device-specific breakpoints like projects usually do) is that different components of the site need to be re-adjusted at entirely different widths i.e because of the dimensions, design variations, hierarchy of the contents and so on.

As a rule of thumb What I personally do is write a breakpoint where a component starts to look bad.

I could literally move that breakpoint to the breakpoints object instead of the component but that I'll end up having a lot of breakpoints that are not being re-used a lot. 

Once again we could define a few device-specific breakpoints in the breakpoints object and then use the nearest breakpoint to whenever things need re-adjustments but this can lead to the site to appear bad on some devices.

Designing for the web today is like designing for an infinite canvas we can never bet on what is gonna be the size of the device. I think we should not be thinking of our web designs in terms of pixels anymore.

The above is my opinion and It could maybe be wrong as I'm not an expert web developer and don't know a lot about responsive Web Design. 

Sorry. I'm not that good at communication please let me know if there is something that I misunderstood or have not explained well.

nisarhassan12: > PS. By the way, what tool do you use to show the dotted line on the webpage, as you did in your previous comment?

I used Firefox dev tools to inspect the element and then clicked on the arrow shown below and then on the actual element. You can also click on the element's markup to have the dotted lines appear.



image

samuelpath:
Thanks for the detailed explanation @nisarhassan12, I appreciate, and I agree with you 👍 .

So I've implemented your suggestion in a separate commit.

Also thanks for the Firefox devtool tip, very helpful.

> Sorry. I'm not that good at communication please let me know if there is something that I misunderstood or have not explained well.

You're doing great 💯

src/pages/blog.tsx
```diff
@@ -33,9 +33,9 @@ const StyledBlogPage = styled.div`
...
```
nisarhassan12, line 36: @samuelpath Many Thanks. Sorry, you have to revert this change i.e:
```suggestion
        justify-content: space-between;
```

other then this we are in a good to merge this.

nisarhassan12:
@samuelpath Once again many thanks for the PR.

PR #741, opened 2020-09-03
[self-hosted] add terms

AlexTugarev:
source: https://github.com/gitpod-io/gitpod/blob/master/License.enterprise.txt


svenefftinge (approved):
LGTM

PR #795, opened 2020-10-08
remove the letter spacing from the tables + adjust the font size for the footnotes

nisarhassan12:


ChristinFrohne (approved):
Great, thanks. 

~~I just noticed another detail, maybe you can quickly address it here as well?
In the first two footnotes it says: "Sep 25". "Sept. 25" would be correct (just like in the footnotes of the table)~~
I just noticed that Jan edited them. I'll discuss it with him

PR #812, opened 2020-10-19
improve icons.

### LINKED ISSUE #802: Improve icon hover style and size
Problem:
- the icon style is inconsistent. Sometimes icons turn dark gray on hover and other times they become blue
Bildschirmfoto 2020-10-12 um 13 00 49

 - the size of the "circle" button slightly differs sometimes
Bildschirmfoto 2020-10-12 um 13 01 08
Bildschirmfoto 2020-10-12 um 13 01 20

Solution:
- let's always use the smaller icon size
- icons should always turn dark gray on hover

nisarhassan12:
Fixes #802

### DIFF
--- a/src/components/Arrow.tsx
+++ b/src/components/Arrow.tsx
@@ -8,7 +8,7 @@ const StyledArrow = styled.svg`
     display: inline-block;
     fill: ${colors.offWhite1};
     transition: all .2s;
-    height: 4rem;
+    height: 4.5rem;
 `
 
 interface ArrrowProps {
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -62,11 +62,11 @@ const StyledFooter = styled.footer`
             color: ${colors.link};
 
             svg {
-                fill: ${colors.link};
+                fill: ${colors.text};
             }
 
             .mail {
-                stroke: ${colors.link};
+                stroke: ${colors.text};
             }
         }
 
--- a/src/components/careers/ExpandableJob.tsx
+++ b/src/components/careers/ExpandableJob.tsx
@@ -80,7 +80,7 @@ const ExpandableJob = ({ title, intro, paragraphs, lists, textAfterTheLists, ren
         <StyledExpandableJob>
             <button onClick={toggleIsRendered}>
                <Arrow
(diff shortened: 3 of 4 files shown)


nisarhassan12:
@ChristinFrohne As we had this discussion in the call that we would like to have all of the icons turn into dark grey on hover. 

Currently for icons that are links we are using the blue color when they are hovered i.e this sort of better indicate that it's a link rather then a button. Should I change to have dark grey on hover is it more apt to have them in blue? Thanks 


image

ChristinFrohne:
Aha that's why they are blue :D 
Interesting thought but I don't think that the users notice that. We're also using the blue color for buttons that don't link to another page. So please use the dark gray color :)

nisarhassan12:
Thanks, @ChristinFrohne I have made the hover styles consistent for the icons (I hope I have not missed any).

Regarding the arrow circle size, I have made it consistent in both places. I have not decreased the size as you said in the issue due to accessibility concerns i.e small tap targets are not a good idea. While I could have made them smaller atleast on large screens like Laptops and so on but touch devices size ranges overlap size ranges of Laptops and so on.

> let's always use the smaller icon size

The Icon size is still in-consistent. Would be great if you explain to me why having the same small icon size is a good idea :) Thanks

ChristinFrohne:
>The Icon size is still in-consistent. Would be great if you explain to me why having the same small icon size is a good idea :) 

The circle-button should always have the same size for consistency reasons. If we decide to make one smaller or bigger there needs to be a significant difference. Minor differences mostly look like mistakes/slappy work. 

Now all the sizes are accurate. Thanks!

Just one minor thing:

- [ ] could you add the hover style to the drop down arrow on /careers as well?

nisarhassan12:
> - [x]  could you add the hover style to the drop down arrow on /careers as well?

Done. Actually there was already the same hover styles on the arrow but not on the acitve/open state i.e the background for the open state was the same as on hover, which is now removed.

ChristinFrohne (approved):
Thanks!
I noticed that the "share icons" on /blog don't have a hover yet. But I am not sure if you're already addressing that in another PR. If so, ignore my comment :) 

Bildschirmfoto 2020-11-03 um 09 57 01

nisarhassan12:
Thanks that is addressed in https://github.com/gitpod-io/website/pull/808#issuecomment-720986176

Issue #821, opened 2020-10-26
SEO improvements


ChristinFrohne:
- [ ] add robots.txt (This file can protect private content from appearing online, save bandwidth, and lower load time on your server. A missing "robots.txt" file also generates additional errors in your apache log whenever robots request one.)

- [ ] apparently there are 5 missing alt tags (e.g. the image next to `where teams code together``

- [ ]  This website loads 9 javascript files. This may affect the load time negatively. Can we reduce them? 

this is the updated version of #561

nisarhassan12:
@ChristinFrohne We can significantly reduce the amount of JavaScript if we were to remove the Twitter embeds i.e all the tweets seem to load their own separate (non-blocking)JS files which sums up to `31.70 MB` which is insane. 

I know we could not get rid of the tweets but maybe instead of the twitter embeds we can re-create them ourself that way we will be able to get rid of `31.70 MB` of JavaScript being shipped ? 

As a comparsion `www.gitpod.io/` i.e the homepage ships 7 JavaScript files summing up to `456.13KB` which is less then `0.5 MB`.

nisarhassan12:
Doc search is a feature that is available for `/docs` not sure why the CSS and JavaScript for the search which is added via Algolia doc search are loaded on every page.

nisarhassan12:
Currently we are serving the fonts from `Google Fonts` would be better to serve them from our own CDN alongside other resources. I'll make a Pr for that.

nisarhassan12:
There are some issues related to image performance.  https://www.gatsbyjs.com/plugins/gatsby-image/ Can help us address the below issues. @ChristinFrohne Should I make a Pr for it ? 

![image](https://user-images.githubusercontent.com/46004116/97335818-17ea5800-18a0-11eb-829a-6edcfda94398.png)

![image](https://user-images.githubusercontent.com/46004116/97335854-22a4ed00-18a0-11eb-9dc9-95f393e8488e.png)

ChristinFrohne:
Thanks for looking deeper into this @nisarhassan12 
>we can re-create them ourself that way we will be able to get rid of 31.70 MB of JavaScript being shipped ?

Would the size be significant smaller then? If so, it seems like a great solution. Then we could also adjust the style of the card and add some information to it. (e.g. Ben Halpern, Founder of dev.to)

>Doc search is a feature that is available for /docs not sure why the CSS and JavaScript for the search which is added via Algolia doc search are loaded on every page.

[...]

PR #823, opened 2020-10-27
self host fonts for better performance.

nisarhassan12:
Addresses https://github.com/gitpod-io/website/issues/821#issuecomment-717348006


nisarhassan12:
Before merging this it should be tested whether the loading performance is better or not i.e via a banchmark comparing this PRs preview with the production site.

ChristinFrohne:
according to Google PageSpeed there is a small speed improvement. Is this the impact that you expected @nisarhassan12 ?

Bildschirmfoto 2020-11-16 um 14 30 55

nisarhassan12:
@ChristinFrohne Yeah. I was expecting some improvement but not this much. On the deploy preview, the fonts don't seem to be loading at all maybe that is the case in the score improvement. It was not the case on the development preview. Do the fonts load for you on the deploy preview ? Thanks

nisarhassan12:
I have fixed the issue due to which the fonts were not loading. Someone should re-review this.

gtsiolis:
Looking at this now! 👀

ChristinFrohne:
Not sure how I can test it as the preview still uses the old embedded tweets and it's therefore a lot slower. Could you adjust the PR to the current state or do you have another idea how to test the performance?

nisarhassan12:
Thanks. @ChristinFrohne I have updated the Pr to the latest state. Please use Google Page Speed + test manually via disabling the cache both versions to see notable differences.

gtsiolis (approved):
Hey @nisarhassan12! Changes LGTM.

However, have you tried using [`fontsource`](https://github.com/fontsource)? The font used (Montserrat) is already available there and we could easily add this by running `npm install fontsource-montserrat`.

This is also one of the preferred ways for adding web fonts with Gatsby (see [relevant guide](https://www.gatsbyjs.com/docs/using-web-fonts/)) and `woff`+`woff2` font files can provide quite sufficient browser support.

Let me know what you think. We can also add a follow-up issue to discuss this or create a new PR if we want to try this. 🏀

nisarhassan12:
Thanks, @gtsiolis.

> However, have you tried using fontsource? The font used (Montserrat) is already available there and we could easily add this by running npm install fontsource-montserrat.

No, I haven't would definitely be something worth trying.

I'm a bit unsure regarding the self-hosted fonts or whether they make a positive impact on performance i.e see the below i.e in the before version the fonts seem to get loaded and rendered faster than the version with self-hosted self-hosted fonts which isn't What I was expecting. Can you also re-produce this? Thanks

|Before|After|
|-|-|
| Peek 2020-12-09 20-10 | Peek 2020-12-09 20-05 |

PR #886, opened 2020-12-08
[dev] Add a delay between port 8000 served & previewed, to give Gatsby time to wake up and be ready

jankeromnes:


jankeromnes:
Does not solve the preview problem 🙄 

Screenshot 2020-12-08 at 16 02 52

PR #904, opened 2020-12-13
Add link to Drupal starterkit.

bserem:
PR adds a link to a Drupal starterkit with MySQL, something which I believe was missing.

The starterkit utilizes Composer 2, Drupal (latest stable) and Drush with MySQL and is suitable for demoing Gitpod to Drupal user and for Drupal development.

It can be used as a base for new projects or as an inspiration for teams to add gitpod in their workflow.


gtsiolis:
Thanks for contributing @bserem! We're currently considering of restructuring this part of the docs (_Example Repositories_) which could be replaced eventually. We'll get back to this as soon as possible.

bserem:
Thanks @gtsiolis, I will be looking to your documentation pages often in any case, so I will catch the update.

ghuntley (approved):
Let's merge this as-is and figure out the documentation restructure in the near future. Thanks for the pull-request @bserem. LGTM.

PR #907, opened 2020-12-16
Video review

vfarcic:
Please let me know whether it makes sense to add the video to the docs and, if it does, where would be a good place


JohannesLandgraf:
Hi @vfarcic, thanks for your PR and the great explainer video about Gitpod 🙇‍♂️ Seems that you fully understood what we are up to! As we want to keep our first intro page in the docs as short as possible we decided to close that PR. For now, I cannot think about another place in our docs where it fits. Once that changes I'll reach out! 🤽‍♂️

Issue #882, opened 2020-12-08
Add a link to the GitHub repository on www.gitpod.io


gtsiolis:
### Problem

We want to add a link to the GitHub repository for the `gitpod-io/gitpod` project on www.gitpod.io.

See relevant discussion in https://github.com/gitpod-io/website/pull/862#issuecomment-740449229.

### Proposal

**Plan A / Option A**: The first MVC (minimum viable change) could include only adding a static link as described in https://github.com/gitpod-io/website/pull/862#issuecomment-740449229:

> If there isn't an easy fix for the dynamically created GitHub star reference, let's start with a static link

**Plan A / Option B**: The second iteration could include a dynamic link which contains the number of stars of the `gitpod-io/gitpod` project on GH.

---

For future reference, I'm pasting below the plan from https://github.com/gitpod-io/website/pull/862#issuecomment-739907446.

> ### Plan A

> We could have an extra item for the GitHub repository but not sure if the current design (font family, spacing, etc) helps.

> The design proposal also changes the top right action to a primary action button which would allow adding the secondary button for GitHub with no hierarchy conflict, see https://github.com/gitpod-io/website/issues/876 for more context.

> | Option A | Option B |
> |-|-|
> | ![image](https://user-images.githubusercontent.com/120486/101353699-b05ef980-389c-11eb-868d-524f8a2f5729.png) | ![image](https://user-images.githubusercontent.com/120486/101353701-b1902680-389c-11eb-81e1-de05696304cb.png) |

> ### Plan B

> Alternatively, we could introduce a secondary button on the body to link to the repository as described in https://github.com/gitpod-io/website/issues/876, which I would strongly recommend regardless if we use this secondary button for linking to the repository.

> To demonstrate this in action, here's how this could look like.

> We could also skip the GitHub icon or rephrase  the button label to _View on GitHub_.

> | BEFORE | AFTER |
> |-|-|
> | ![image](https://user-images.githubusercontent.com/120486/101350356-82c38180-3897-11eb-8d3f-0aa9e8a6c344.png) | ![image](https://user-images.githubusercontent.com/120486/101350381-8951f900-3897-11eb-9b2a-8a63cf731aef.png) |

ChristinFrohne:
I'm a bit afraid that having two buttons in the navigation bar might be confusing and visually a bit too much/ too heavy. If we have several highlighted options to click on the users might be confused and don't click anywhere.

[...]

Issue #876, opened 2020-12-07
Improve header navigation


gtsiolis:
### Problem

1. We're missing a CTA button on the header navigation.
2. We want to add a link to the GitHub repository, per https://github.com/gitpod-io/website/pull/862#issuecomment-739506303.

### Proposal

1. Replace the top right sign in button with a primary action button.
1. Introduce a secondary button next to the primary button the body.
1. Use the new secondary button to link to GitHub.

To make the proposal more clear, here're some mockups.

| BEFORE | AFTER |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/101349233-bbfaf200-3895-11eb-862d-d1c01dd5c0ff.png) | ![image](https://user-images.githubusercontent.com/120486/101349237-bc938880-3895-11eb-9540-788b55f16564.png) |

ChristinFrohne:
I like the idea of having a CTA button in the navigation bar. When we first created the website the thought was:
- have a `login` button if you already have an account
- have a `get started`button if you're new to Gitpod

Unfortunately we'd couldn't get that implemented. What do you think about this? Any idea how to do that? Maybe by using cookies?

The`get started button` could link to https://www.gitpod.io/#get-started

PR #911, opened 2020-12-17
Add getting started button

gtsiolis:
This will add a secondary button to link to the new [Getting Started](https://www.gitpod.io/docs/getting-started/) section in the docs which mentions how you can use Gitpod.

**Context**: This builds on top of the original proposal in https://github.com/gitpod-io/website/issues/882#issue-759375835 and moves us one step closer to https://github.com/gitpod-io/website/issues/876.

| BEFORE | AFTER |
|-|-|
| image | image |

### DIFF
--- a/src/components/index/Banner.tsx
+++ b/src/components/index/Banner.tsx
@@ -94,6 +94,9 @@ const Banner = () => (
                 <a href="#get-started" className="btn btn--cta">
                   Try Now
                 </a>
+                <a href="https://www.gitpod.io/docs/getting-started/" className="btn" style={{ marginLeft: '8px' }}>
+                    Getting Started
+                </a>
                 <p className="works-with">
                   Works with
                   <img src={GitLab} alt="GitLab" />


gtsiolis:
@JohannesLandgraf WDYT? Feel free to close this if you think this is going into the wrong direction. /cc @ChristinFrohne

src/components/index/Banner.tsx
```diff
@@ -78,6 +78,9 @@ const Banner = () => (
             <a href="#get-started" className="btn btn--cta">
               Try Now
             </a>
+            <a href="https://www.gitpod.io/docs/getting-started/" className="btn" style={{ marginLeft: '8px' }}>
```
gtsiolis, line 81: **thought**: Alternatively, we could link to the [Introduction](https://www.gitpod.io/docs/) section, it just seemed natural and more practical to link to the [Getting Started](https://www.gitpod.io/docs/getting-started/) section. /cc @JohannesLandgraf

ChristinFrohne:
I as a user would be confused about the differences between `try now` and `get started`. Instead, we could add the link to the documentation to https://www.gitpod.io/#get-started

src/components/index/Banner.tsx
JohannesLandgraf: Agree, also the Getting Started Section links back to the Introduction Section so should be fine.

JohannesLandgraf (approved):
I really like the idea @gtsiolis, let's merge after you resolved the conflicts

gtsiolis:
Merging per https://github.com/gitpod-io/website/pull/911#pullrequestreview-582184591. Thanks @JohannesLandgraf @ChristinFrohne for the feedback! 🙏 

@ChristinFrohne regarding https://github.com/gitpod-io/website/pull/911#issuecomment-747457299, this will probably become better and more clear after the redesign but I'd suggest to avoid removing the only primary action that prompts the user to immediately open a new environment using an example project from this section. Let's discuss this in a new issue if you think it's worth putting more effort in this before the redesign. 🌱

Issue #905, opened 2020-12-15
Make the editing docs button more visible and prominent


gtsiolis:
### Problem to solve

Currently, the _Edit this page in Gitpod_ button on the docs is not making it clear that this will _**not**_ edit the markdown file in GitHub as commonly seen elsewhere.

There's a lot of magic going on behind a click on that pencil button.

### Proposal

Let's make this button more visible and prominent.

This could use the _Open in Gitpod_ button as seen in various repositories or create a new one with the label _Edit in Gitpod_ if needed.

| BEFORE | AFTER |
|-|-|
| ![image](https://user-images.githubusercontent.com/120486/102217428-3a364480-3ee5-11eb-9eca-602c1ed3c729.png) | ![image](https://user-images.githubusercontent.com/120486/102217435-3c000800-3ee5-11eb-8b5f-c0491a0f3358.png) |

ChristinFrohne:
I like your idea of having an `edit in Gitpod` button. Just using the `open in Gitpod` button wouldn't convey the message that everyone can contribute and complement the documentation.  
Alternatively we could name it `contribute in Gitpod`

gtsiolis:
The verb _Edit_ is also common across a lot of documentation tools. _Contribute_ could be also nice.

| VuePress | Docusaurus | Read the Docs | docsify | ... |
|-|-|-|-|-|
| ![image](https://user-images.githubusercontent.com/120486/102342617-2e5c8800-3fa2-11eb-808a-9e499e0c822f.png) | ![image](https://user-images.githubusercontent.com/120486/102342731-551abe80-3fa2-11eb-8e1e-cba834aa84bc.png) | ![image](https://user-images.githubusercontent.com/120486/102342741-56e48200-3fa2-11eb-84b8-68f3307382c2.png) | ![image](https://user-images.githubusercontent.com/120486/102342750-58ae4580-3fa2-11eb-9f8e-25ddff54d912.png) | ... |

We could also use a simple link button on every page.

ChristinFrohne:
Works both for me. The Gitpod button would be more visible and stick out more though. And it might also help to educate the user: every time they see the blue button, they know that it will start a complete dev environment

JohannesLandgraf:
+1 for @gtsiolis suggestion of the blue Gitpod button with the label "Edit in Gitpod", which is more descriptive for docs than "Contribute" - aaand another 0.5% improvement 🚀

nisarhassan12:
[...]

PR #918, opened 2020-12-28
Make the editing docs button more visible and prominent.

nisarhassan12:
Fixes gitpod-io/website#905

This is how it looks:

image


ChristinFrohne:
Thanks!

- [ ] please align the button to the right so that its on the same height as the text?
- [ ] could you add a hover? A lighter blue like #5093EF would be great

nisarhassan12:
> - [ ] please align the button to the right so that it's on the same height as the text?

I have aligned the button to the left with the navigation(though it is not perfectly aligned across all screen sizes). Having it on the same height as the text is doable but it is not so simple to achieve. Is it something important to do? Thanks

ChristinFrohne:
Sorry, I think I haven't expressed myself clearly. 

I meant, aligning the button to the right so that it stops with the text. Currently, there is a small gap. Can we simply move the button a bit to the right?
Bildschirmfoto 2021-01-04 um 16 47 24

nisarhassan12:
@ChristinFrohne This is how it looks now:

image

src/components/docs/EditInGitpod.tsx
```diff
@@ -1,41 +1,61 @@
...
-        border-radius: 50%;
-        transition: all .2s;
-
-        &:hover,
-        &:focus {
-            background: ${colors.text};
+        &:hover {
+            rect {
+                transition: all .2s;
+                fill: #5093EF;
```
ChristinFrohne, line 14: I adjusted the color as it was a bit too bright

ChristinFrohne (commented):
I adjusted the colors for the hover effect, as it was too bright

nisarhassan12:
> I adjusted the colors for the hover effect, as it was too bright

Great. Thanks. Can we merge this now?

PR #920, opened 2020-12-30
[docs]: categorize frameworks in submenus under their relevant language 

nisarhassan12:
Fixes gitpod-io/website#783

This is how it looks:

image

### DIFF
--- a/src/components/Linkset.tsx
+++ b/src/components/Linkset.tsx
@@ -4,7 +4,7 @@ import styled from '@emotion/styled'
 import { colors } from '../styles/variables'
 import { Link } from 'gatsby'
 
-const StyledLinkSet = styled.div`
+const StyledLinkSet = styled.ul`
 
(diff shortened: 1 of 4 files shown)


ChristinFrohne:
I think adding a submenu is not so relevant right now, as there are barely any items in the submenu, and it creates confusion in the navigation overview. 
I'd suggest leaving it as it is for now and introducing another navigation structure at a later point when we have more frameworks to list. 

If you think it's necessary to change now, let me know, and I will look into how to structure the navigation bar better.

gtsiolis:
Agree with @ChristinFrohne. Introducing new structure, table of contents, or more could be more useful as the content grows. However, at the moment this change could cause for confusion. I'd suggest closing this for now. ❤️

nisarhassan12:
Many Thanks, @ChristinFrohne @gtsiolis for your feedback.

Issue #842, opened 2020-11-24
SEO Improvements


ChristinFrohne:
**Page not found:** 

- [ ]  https://www.gitpod.io/docs/languages-and-frameworks/ links to https://www.gitpod.io/docs/frameworks/pandas/ ->404

- [ ] https://www.gitpod.io/docs/gitlab-integration/ and https://www.gitpod.io/docs/bitbucket-integration/ link to https://www.gitpod.io/access-control/ ->404

- [ ] https://www.gitpod.io/docs/languages/dart/ links to https://gitpod.io/docs/config-docker -> 404

- https://www.gitpod.io/docs/languages-and-frameworks/ links to https://www.gitpod.io/docs/kotlin-in-gitpod/ -> 404

- [ ] https://www.gitpod.io/docs/languages/java/ links to https://www.gitpod.io/docs/vscode_extensions/ -> 404

**Page not found -external links**
- [ ] https://www.gitpod.io/blog/gitpodify/ to -> 404 https://github.com/Chocobozzz/PeerTube/blob/master/scripts/watch/client.sh

- [ ] https://www.gitpod.io/blog/build-a-webhook-for-google-assistant/ links to https://developers.google.com/actions/samples/github -> 404

**Missing alt-tag**

- [ ] https://www.gitpod.io/static/vendor-bg-fe98fcdf8bdc08026e8d2d390dab1f5d.png   

- [ ] https://www.gitpod.io/static/college-students-03ccf407c383bd4bcc1ef7dd2d5a78ec.png      

- [ ] https://www.gitpod.io/static/recruiting-bg-5d4ca5b37cedee043d7d79a6dca6ca44.png      

**meta description is too short**
Google only crawls pages that have a certain length of meta description. 60-160 characters are recommended
- [ ] all pages under /doc
Current meta description `Documentation site for Gitpod.` 
Suggestion: `Explore the documentation to learn more about Gitpod.io and Gitpod Self-Hosted.`

- [ ] /blog 
current meta description: `Discover articles and tutorials about Gitpod`
suggestion: `Visit the Gitpod blog to learn about releases, tutorials, news and more`
- [ ] meta description for gitpod.io needs to be updated. Currently `One-click ready-to-code development environments for GitHub`. New suggestion: `Gitpod is an open source developer platform automating the provisioning of ready-to-code development environments."

Generally we should watch out that the meta description for new blog posts is not too short

Search Engine Result Pages (SERP)
[...]

PR #932, opened 2021-01-05
make site valid in accordance to W3C validator.

nisarhassan12:
Addresses the  following from gitpod-io/website#842

image


nisarhassan12:
Not sure How we can get rid of this error:
image

nisarhassan12:
According to W3C we should avoid nesting Links:

image

But currently, we're doing it on https://www.gitpod.io/blog in the PostPreview Card

image

Removing the links from the author name would fix this issue. @ChristinFrohne Is this ok to remove the links from the names? Thanks

image

ChristinFrohne:
@nisarhassan12 I just noticed that the links to the author's profile are not working anyways. So you can just remove the link completely from the overview page. On each blog post the link to the author's profile needs to be there though.

PR #958, opened 2021-01-20
add screencasts. 

nisarhassan12:
This is how it looks so far:

image

image


ChristinFrohne:
Thanks for the superfast implementation @nisarhassan12 

**Question:** 
- Why did you decide to use the loom player instead of embedding the YouTube video? If we use the embedded YouTube video, the user might follow us on YouTube and discovers other videos about Gitpod or get notifications once we publish new videos. Currently, there is also a short preview of a dog before the player is loaded ❓
- How can the user access the page? We definitely need to add it to the footer. 

**Suggestions:**
- I personally favor to have the time written in minutes as soon it extends 60secs. I find it more explicit and time-saving
- we need to link the content well through the website. It would be helpful to add a link to some features. For the documentation it might be best to embed the related screencast directly. 
- please align the "next up" video in the center of the screen 
- there can be a bit more spacing between the player and the font. For example here
Bildschirmfoto 2021-01-22 um 09 23 01

/cc @jankeromnes

ChristinFrohne:
Here is the summary from the Slack discussion:

- [ ] let's embed the YouTube videos (when uploading them to YouTube, we'll make sure to disable video recommendations to eliminate distracting noise)

- [ ] please add the link to the screencasts to the footer under "developer". After this PR is merged, we'll open another PR to add the screencasts to the navigation bar. @gtsiolis will suggest a layout for that

src/components/screencasts/NextUp.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface NextUpProps {
+    nextScreencast: ScreencastProps | null
+}
+
+const NextUp = ({ nextScreencast }: NextUpProps) => (
+    nextScreencast ? (
+        <div className="row">
+            <StyledNextUp>
+                <h2>Next Up ...</h2>
+                <Screencast headingType="h3" {...nextScreencast}/>
```
gtsiolis, line 18: **suggestion**: What do you think of aligning this to the center as @ChristinFrohne mentioned in https://github.com/gitpod-io/website/pull/958#issuecomment-765229315?

src/components/screencasts/RelatedDocs.tsx
```diff
@@ -0,0 +1,31 @@
...
+    path: string;
+}
+
+const StyledRelatedDocs = styled.div`
+    margin-bottom: 10rem;
+`
+
+const RelatedDocs = ({ relatedDocs }: {relatedDocs: RelatedDoc[]}) => (
+    <StyledRelatedDocs>
+        Related Documentation:&nbsp;
```
gtsiolis, line 17: **praise**: Nice touch adding this below!

src/contents/screencasts.json
```diff
@@ -0,0 +1,52 @@
+{
+    "screencasts": [
+        {
+        "title": "Always ready-to-code",
+        "description": "Lorem ipsum dolor sit amet consectetur adipisicing elit. Illum minima quas laboriosam, totam officiis magnam voluptate quibusdam eius harum molestiae necessitatibus.",
```
gtsiolis, line 5: **nitpick**: Let's replace these with some actual text or just remove all of them. Both options would suffice.

src/components/screencasts/Screencast.tsx
```diff
@@ -0,0 +1,76 @@
...
+const StyledScreencast = styled(Link)`
+    display: block;
+    padding: 2rem;
+    text-align: center;
+    border-radius: 3px;
+    max-width: 500px;
+    min-width: 400px;
+    border: 1px solid #eee;
+    color: ${colors.text};
+    transition: all .2s;
```
gtsiolis, line 20: **nitpick**: Feel free to leave this as is for now but I'm not sure if a) the transition duration and b) applying this to two rows of text  is optimal here although looks fun. 🌈

gtsiolis (commented):
Looks great @nisarhassan12! Adding some minor comments below.

Also agree with @ChristinFrohne to add this to the footer under the _Developer_ section.

JohannesLandgraf:
> Also agree with @ChristinFrohne to add this to the footer under the Developer section.

As long we are not really educating users within the product we should make it as easy as possible to consume content that helps them to understand what Gitpod is about. Hence screencasts should be more prominently surfaced on the website (and also added/linked in the documentation). 

Wdyt about implementing the centralized top-line navigation for now (similar to https://buildkite.com/) you had already some designs for @gtsiolis? We should also not overcomplicate things and could stay quite close to the Buildklite navigation bar

Left-side links: Features, Screencasts, Docs 
Right-side: Community, Pricing, Install

src/components/screencasts/NextUp.tsx
jankeromnes: Done 😇

src/contents/screencasts.json
jankeromnes: Added official descriptions from YouTube 🙌

jankeromnes:
> As long we are not really educating users within the product we should make it as easy as possible to consume content that helps them to understand what Gitpod is about. Hence screencasts should be more prominently surfaced on the website (and also added/linked in the documentation).

100% agree! ☝️

For now I've added a link in the footer, but let's also add a prominent link in the top menu bar right after this PR is merged (as suggested by @ChristinFrohne in https://github.com/gitpod-io/website/pull/958#issuecomment-765245442).

jankeromnes:
I think this PR should be ready-to-go now. Just waiting on Netlify to wake up and do the thing.

gtsiolis:
> Hence screencasts should be more prominently surfaced ...

@JohannesLandgraf Agree! Adding the item on the footer was only suggested as a complementary change.

jankeromnes:
I'll go ahead and merge this. Next let's add a prominent Screencasts link at the top of the website.

(Also, I've noticed a small problem with spacing around the video, that fixes itself after a refresh -- hopefully that's just a caching bug, but if not we may want to fix it before making the page more prominent.)

PR #959, opened 2021-01-21
Added docs for prebuild branch filtering on Gitlab

j0nes2k:
By default, the Gitlab webhook that is installed by Gitpod triggers prebuilds on any push event on any branch. This behavior can be changed in the Gitlab settings, for example to trigger prebuilds only on master branch.


ghuntley:
Thanks for your contribution @j0nes2k, I didn't have permission to push to your branch (this can be enabled in GitHub settings) so work has continued over at https://github.com/gitpod-io/website/pull/1003 which includes your contribution.

Issue #963, opened 2021-01-25
Create deeplinks for each of the job openings


JohannesLandgraf:
We should have deeplinks for each individual job opening to allow for easier distribution

PR #964, opened 2021-01-25
Create deeplinks for each of the job openings.

nisarhassan12:
Fixes gitpod-io/website#963

I have hyphenated the titles in lowercase to be the hashes belonging to each of the sections i.e this is how the deeplinks are gonna look like for each of the job post:
#835 

- www.gitpod.io/careers/#software-engineer
- www.gitpod.io/careers/#senior-software-engineer-(full-stack)
- www.gitpod.io/careers/#senior-software-engineer-(workspace-runtime)
- www.gitpod.io/careers/#senior-devops-engineer

### DIFF
--- a/src/components/careers/ExpandableJob.tsx
+++ b/src/components/careers/ExpandableJob.tsx
@@ -1,9 +1,11 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 
 import styled from '@emotion/styled'
 import { colors, borders } from '../../styles/variables'
 import Arrow from '../Arrow'
 import GitpodLogo from '../../resources/gitpod-logo-dark.svg'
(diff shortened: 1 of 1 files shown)


meysholdt:
Nice! 

As a person who would post the links on the internet, can you give me an easy way to copy the deep-link from the website? 
If there is one already and I just overlooked it, let me know.

nisarhassan12:
Thanks, @meysholdt there was not any way to copy the deep link for sharing easily so I have added one:

Peek 2021-01-28 00-35

It is almost very similar to what we have on the docs. We could go more further and provide a copy to clipboard button which copies the deep-link to the clipboard with a single click What do you say ? Thanks

JohannesLandgraf:
Looks good - I'll merge

Issue #967, opened 2021-01-26
Include a banner on top of our website that we are hiring


JohannesLandgraf:
We should drive more traffic on our hiring page as possible to attract the best talent. As a positive side effect that banner also indicates that the company is growing. We can take the banner we used for going open source as Vorlage. Alternatively use a single-colored (#0087BE) banner which says: "We are hiring! 🌈 " and with a small circle indicating the number of openings (similar to https://raycast.com/ does in their top navigation bar).

@ChristinFrohne Could you come up with a rough sketch/design you like?
@nisarhassan12 You can already start and  implement logic etc in line with the open source banner or a design you think looks sweet 🍭

ChristinFrohne:
I'd have the banner a bit more narrow than the one we had for the open-source launch. It looked a bit bulky to me after all. I like how raycast is indicating the amount of open position, but without `jobs` the number could be confusing.

It could look like this:
<img width="636" alt="Bildschirmfoto 2021-01-26 um 15 43 21" src="https://user-images.githubusercontent.com/46345125/105859823-3cfc7d00-5fed-11eb-9592-94e1ad95174a.png">

JohannesLandgraf:
I would make the banner even narrower (especially if we use a single colour for it) - otherwise looks great! But it looks like the colour is a bit lighter than the button (we should probably take the same one from the button)

ChristinFrohne:
It's an optical illusion. The color of the banner and the "try now" button have the same color. The color of "4 Openings" is the same as we're using for hover effects.

JohannesLandgraf:
Let's also add the 🌎 after the rainbow to indicate that we are working remotely

nisarhassan12:
@JohannesLandgraf I'm working on it now will hopefully finish it before 3:00 PM your time.

JohannesLandgraf:
Hm.. not sure what I am doing wrong but I don't see a banner (using Brave with uOrigin)

nisarhassan12:
@JohannesLandgraf that is expected i.e once you click the close icon or click the opening button we remove the banner forever just like it was the case for the Gitpod Open Source Announcement. Should remove the close icon and make the banner display forever on all pages? Thanks

nisarhassan12:
I have removed the closed button and made the Banner always visible across all pages.

PR #972, opened 2021-01-27
add announcement banner that we are hiring.

nisarhassan12:
Fixes gitpod-io/website#967

This is how it looks: 

image


JohannesLandgraf:
Thanks @nisarhassan12! Seems to work well. However it is way too invasive and prominent. The banner should probably be at most half the size and the font size also respectively smaller. There is also no need to use bold font here. To be a bit more colloquial I would change the text to: "We're hiring".

nisarhassan12:
I have made the requested changes.

JohannesLandgraf:
Thanks @nisarhassan12! 

- The button to close the banner is not visible anymore
- We should resize the opening hover button. `font-size`:1.4rem / `margin-left`: 1.5rem /
- Change the text to "We're hiring!"

Such as:
Screenshot 2021-01-27 at 18 34 06
 
 Maybe even decrease the banner height even more (but up to you to decide what still looks good) :)

JohannesLandgraf (approved):
Once you fix the things in my comment above feel free to merge on your own 🙏

Issue #977, opened 2021-01-28
Job opening banner isn't visible in Chrome


ChristinFrohne:
**Problem**
On Chrome: The banner is only blinking up very shortly and is then closed automatically. 
On Safari: works fine
On Firefox: works fine

**Additional Feedback**
- Let's not use a button style for "8 openings" but rather link the whole banner to the career page
- and we can remove the closing icon. As the banner isn't fixed while scrolling, it doesn't take much space and can be visible at all time.  

Could you look into it @nisarhassan12 ?

nisarhassan12:
Thanks, @ChristinFrohne for reporting this. Although I was not able to reproduce this on Chrome I have attempted a fix in https://github.com/gitpod-io/website/pull/980 Please have a look at the preview you'll hopefully not see the issue there. 

![image](https://user-images.githubusercontent.com/46004116/106310430-5d466900-6285-11eb-8da1-8dba318986cf.png)

> we can remove the closing icon. As the banner isn't fixed while scrolling, it doesn't take much space and can be visible at all time.

Maybe it's just me who thinks like this but I would say it is good to keep that close button on the banner which permanently removes it forever. https://www.pulumi.com/ does it similarly. I think one might not call the Banner an ad but it sort of slightly falls into the category off not being relevant to the core contents just like Ads. I think it is better to give users the ability to get rid of it if they want to? I'm not too sure. What do you say? Thanks

ChristinFrohne:
@nisarhassan12 Currently, the CTA to close the banner is more prominent than the link to the job openings. I'm ok with keeping a closing option, but it should be less prominent.

Could you:
- [ ] not use a button style for "8 openings" but rather link the whole banner to the career page (see www.netlify.com as an example)
- [ ] change the exit-button style to a simply cross
- [ ] have the banner a bit more narrow


Mockup:
<img width="961" alt="Bildschirmfoto 2021-02-01 um 14 25 56" src="https://user-images.githubusercontent.com/46345125/106464832-7a538580-6499-11eb-8678-137c383eceeb.png">

PR #980, opened 2021-01-29
fix: Job opening banner isn't visible issue on Chrome.

nisarhassan12:
Fixes gitpod-io/website#977


ChristinFrohne (approved):
It's visible now. Thanks!

PR #984, opened 2021-02-01
Improve Design and Accessibility for the Banner. 

nisarhassan12:
Fixes gitpod-io/website#977 #992 

This is how it looks:

image

### DIFF
--- a/src/components/AnnouncementBanner.tsx
+++ b/src/components/AnnouncementBanner.tsx
@@ -4,96 +4,102 @@ import { Link } from 'gatsby'
 import { colors, borders } from '../styles/variables'
 import { jobs } from '../pages/careers'
 
-const StyledAnnouncementBanner = styled.div`
-  background: ${colors.link};
-  color: ${colors.white};
-  border-bottom: ${borders.light};
-  transition: all 0.3s;
+const AnnoucementBannerWrapper = styled.div`
+    position: relative;
+    background: ${colors.grey};
+    transition: all 0.3s;
 
-  @media (max-width: 1100px) {
-    background-position: bottom;
+    button {
+        position: absolute;
+        top: 50%;
+        transform: translateY(-50%);
+        right: 2rem;
+        border: none;
+        border-radius: 50%;
+
+        @media(max-width: 472px) {
+            right: 1rem;
+        }
+
+        @media(max-width: 320px) {
+            right: .5rem;
+        }
+    }
+
+    .cross {
+        display: block;
+        height: 1.5rem;
+        width: 1.5rem;
+        fill: ${colors.white}
+    }
+`
+
+const StyledAnnouncementBanner = styled(Link)`
+  color: ${colors.white};
+  font-weight: 400;
+  border-bottom: ${borders.light};
+
+  &:hover,
+  &:focus {
+    color: ${colors.white};
   }
 
   @media (max-width: 600px) {
     font-size: 85%;
   }
 
+  .row {
+    @media(max-width: 472px) {
+        padding: 0 2.8rem 0 .5rem;
+    }
+  }
+
   .text {
-    position: relative;
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    padding: .8rem 0;
+    padding: 1rem 0;
+    font-weight: 600;
+    text-align: center;
+    font-size: 90%;
 
-    @media(max-width: 340px) {
-        justify-content: flex-start;
+    @media(max-width: 472px) {
+        text-align: left;
     }
 
-    @media(max-width: 290px) {
-        justify-content: center;
+    @media(max-width: 410px) {
+        text-align: center;
     }
-  }
 
-  a {
-    font-weight: 400;
-  }
-
-  .wrapper {
-    @media(max-width: 285px) {
-        display: block;
-        margin-bottom: .5rem;
+    @media(max-width: 320px) {
+        padding: .8rem;
     }
   }
 
-  .emojis {
-      margin-left: 1rem;
-      letter-spacing: 3px;
-  }
-
-  .btn {
-    background: ${colors.lightBlue};
-    border-color: ${colors.lightBlue};
-    margin-left: 1.5rem;
-    font-size: 1.4rem;
-
-    @media(max-width: 300px) {
-        font-size: 1.3rem;
-        margin-left: .3rem;
-    }
-
-    &:hover,
-    &:focus {
-        color: ${colors.text};
-        background: ${colors.white};
-        border-color: ${colors.text};
-    }
-  }
-
-  button {
-    position: absolute;
-    right: 0;
+  .openings {
     background: ${colors.white};
-    border: none;
-    border-radius: 50%;
-  }
+    color: ${colors.grey};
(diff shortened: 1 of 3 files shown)


ChristinFrohne (approved):
Looks better, thanks. Personally I prefer having the light blue highlight around the "openings" which was there in the previous design. It should only be a visual highlight without the hover effect. It then reminds me of the issue labeling in GitHub.  
Please merge the option that you find fitting best.

nisarhassan12:
This is how it looks now:

image

src/components/AnnouncementBanner.tsx
```diff
@@ -140,47 +142,36 @@ const AnnoucementBanner = () => {
...
-                            </g>
-                        </svg>
-                    </button>
+                wasDisplayed() ? 'none' :
+                    'inline'
+        }}>
+            <StyledAnnouncementBanner to="/careers/#jobs">
+                <div className="row">
+                    <div className="text">
+                        Wanna help make Gitpod better? join our team!
```
ChristinFrohne, line 153: "Join" needs to be spelled with a capital letter

ChristinFrohne (commented):
I'd really like to get this merged, so I'll just do the review. Feel free to add comments in a follow-up @gtsiolis 

The arrow doesn't seem to be centered. Please align it with the text and add some more spacing to both sides of the button. 
Bildschirmfoto 2021-02-08 um 14 10 12


Feel free to merge it afterwards directly. Thanks!

src/components/AnnouncementBanner.tsx
```diff
@@ -140,47 +146,55 @@ const AnnoucementBanner = () => {
...
+                    'inline'
+        }}>
+            <StyledAnnouncementBanner to="/careers/#jobs">
+                <div className="row">
+                    <div className="text">
+                        Wanna help make Gitpod better? Join our team!
+                        <span className="openings">
+                            <strong>{jobs.length}</strong>
+                            &nbsp;Openings <i>
+                                <svg
```
gtsiolis, line 161: **nitpick**: Pasting here the arrow SVG source in case you would like to make the arrow match the original specs..


```javascript
<svg width="20" height="11" viewBox="0 0 20 11" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M15 1.5L19 5.5M19 5.5L15 9.5M19 5.5H1" stroke="#3D515E" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
</svg>
```

src/components/AnnouncementBanner.tsx
```diff
@@ -140,47 +146,55 @@ const AnnoucementBanner = () => {
...
+                wasDisplayed() ? 'none' :
+                    'inline'
+        }}>
+            <StyledAnnouncementBanner to="/careers/#jobs">
+                <div className="row">
+                    <div className="text">
+                        Wanna help make Gitpod better? Join our team!
+                        <span className="openings">
+                            <strong>{jobs.length}</strong>
+                            &nbsp;Openings <i>
```
gtsiolis, line 160: **issue**: I don't think we need the `<i>` element here, right? The `<i>` element is used for italics, see [relevant docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i).

src/components/AnnouncementBanner.tsx
```diff
@@ -140,47 +146,55 @@ const AnnoucementBanner = () => {
...
-                        }}
-                    >
-                        <svg xmlns="http://www.w3.org/2000/svg" width="29.86" height="29.86" data-name="Banner closing" viewBox="0 0 30.86 30.86">
-                            <g data-name="Gruppe 950">
-                                <g data-name="Pfad 1398">
-                                    <path d="M14.93 29.86a14.93 14.93 0 1114.93-14.93 14.93 14.93 0 01-14.93 14.93z"></path>
+                wasDisplayed() ? 'none' :
+                    'inline'
+        }}>
+            <StyledAnnouncementBanner to="/careers/#jobs">
```
gtsiolis, line 154: **issue:** Was the hyperlink on the row intended? The [button below](https://github.com/gitpod-io/website/pull/984/files#diff-069170cb72394236d5b8b390b155eca52b79cddc59596e82e70aff09f13e393eR158) seems like a perfect item to provide the link to `/careers/#jobs`, right?

src/components/AnnouncementBanner.tsx
```diff
@@ -4,96 +4,102 @@ import { Link } from 'gatsby'
...
-      margin-left: 1rem;
-      letter-spacing: 3px;
-  }
-
-  .btn {
-    background: ${colors.lightBlue};
-    border-color: ${colors.lightBlue};
-    margin-left: 1.5rem;
-    font-size: 1.4rem;
+  .openings {
```
gtsiolis, line 76: **nitpick**: Adding proper _hover_ and _active_ states for the button could make this banner more keyboard accessible. Alternatively, removing the button shape and using a plain text element could help.

src/components/AnnouncementBanner.tsx
```diff
@@ -4,96 +4,102 @@ import { Link } from 'gatsby'
...
   .text {
-    position: relative;
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    padding: .8rem 0;
+    padding: 1rem 0;
+    font-weight: 600;
+    text-align: center;
+    font-size: 90%;
```
gtsiolis, line 61: **issue**: The font-size on small widths (mobile devices) makes the text a bit squeezed and really small. We could remove the button altogether to make room for smaller devices here.

gtsiolis (commented):
Hey @nisarhassan12! Great work!

I didn't get in time to review this but left some comments in case you'd like to tackle these in a separate issue. I'll defer to @ChristinFrohne to decide here.

Happy to discuss about each point raised below.

src/components/AnnouncementBanner.tsx
ChristinFrohne: I agree the current arrow is a bit skinny

src/components/AnnouncementBanner.tsx
ChristinFrohne: The link should be accessible as easy as possible, so I think it's fine to have it on the whole row

src/components/AnnouncementBanner.tsx
ChristinFrohne: Nice point! Makes sense to improve keyboard accessibility. The button style looked bad on the last version. But here it's fine.

src/components/AnnouncementBanner.tsx
ChristinFrohne: I think it looks okay right now. 
Bildschirmfoto 2021-02-09 um 16 24 12

src/components/AnnouncementBanner.tsx
ChristinFrohne: Let's take these specs. Currently, the arrow looks a bit skinny :D

ChristinFrohne:
Thanks for your comments @gtsiolis 
Please implement the remaining issues @nisarhassan12. Thanks! :)

PR #1000, opened 2021-02-12
adjustments banner small improvements.

nisarhassan12:
Addresses the remaining feedback on https://github.com/gitpod-io/website/pull/984

| Before | After |
| - | - |
|image| image|
|image|image|

### DIFF
--- a/src/components/AnnouncementBanner.tsx
+++ b/src/components/AnnouncementBanner.tsx
@@ -17,12 +17,8 @@ const AnnoucementBannerWrapper = styled.div`
         border: none;
         border-radius: 50%;
 
-        @media(max-width: 472px) {
-            right: 1rem;
-        }
-
-        @media(max-width: 320px) {
-            right: .5rem;
+        @media(max-width: 505px) {
+            display: none;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/components/AnnouncementBanner.tsx
```diff
@@ -58,29 +54,23 @@ const StyledAnnouncementBanner = styled(Link)`
...
+    display: inline flex;
+    align-items: center;
     background: ${colors.white};
     color: ${colors.grey};
     padding: .5rem 1.4rem;
     border-radius: 10rem;
     margin-left: 2rem;
 
     @media(max-width: 505px) {
+        display: none;
```
gtsiolis, line 73: **suggestion**: Removing the button with the arrow could make this banner looks non-actionable. The [original thought](https://github.com/gitpod-io/website/pull/984#discussion_r572283143) was to remove the button but keep an inline link. In this case, we could do the following for smaller viewports:
1. Remove the text _Wanna help make Gitpod better?_ and keep only _Join our team!_ in the banner.
2. Optionally, keep the _8 Openings_ part.
3. Keep the arrow icon using white fill but without the white background.
4. Remove the dismiss `X` button.

gtsiolis (commented):
Nice work @nisarhassan12! Left some non-blocking comments.

Since we're going to redesign everything here I'll leave this to @ChristinFrohne to decide if it's ok to merge this as is. 🏀

ChristinFrohne (commented):
Would be great if you can address @gtsiolis's feedback and then merge the PR @nisarhassan12 

We shouldn't spend too much time on this little change

PR #1002, opened 2021-02-14
Fix typo in docs

gragib:
I am pretty sure this should read `GitHub` instead of `GitLab`.


gtsiolis:
You are correct @gragib! Thanks for catching this! 🦊

PR #1006, opened 2021-02-17
[enterprise-license] Remove some unnecessary spaces in contact email subject

jankeromnes:


jankeromnes:
I think this doesn't require a review.

Issue #1009, opened 2021-02-22
[browser extension] uninstall feedback is not being forwarded


ChristinFrohne:
**Problem**
we created a feedback form for users uninstalling the browser extension. The submitted information is currently lost in space. 
<img width="1259" alt="Bildschirmfoto 2021-02-22 um 12 25 59" src="https://user-images.githubusercontent.com/46345125/108702144-23c1f000-7509-11eb-8f21-294b07742f05.png">

**Expected behavior**
send user feedback to contact@gitpod.io. Afterward, we would link the information to the feedback channel in slack

Sidenote:
can we track the user's email address automatically in the background and add it to the feedback as well?

ChristinFrohne:
Could you look into this @ghuntley ?

ghuntley:
Upon uninstallation https://github.com/gitpod-io/browser-extension/commit/154036a9e4c34bd6dd0d3448db1fcdba9ee7f69b#diff-11db236acd6b02b1229e7d325e9bf8edb521032e6388f68b6a884ec24bbe291fR27 opens https://www.gitpod.io/extension-uninstall/ which is https://github.com/gitpod-io/website/blob/master/src/pages/extension-uninstall.tsx

PR #1011, opened 2021-02-26
[browser extension] resolve uninstall feedback not being forwarded

ghuntley:
Resolves https://github.com/gitpod-io/website/issues/1009 when the PR is complete. Opening early for feedback from @gtsiolis because the design has been changed. 

Proposed changes:

1. collect name and email address so the team can respond to learn more
2. add gdpr consent to the collection of information

image

Compare against https://www.gitpod.io/extension-uninstall/

### DIFF
--- a/src/components/extension-uninstall/Form.tsx
+++ b/src/components/extension-uninstall/Form.tsx
@@ -48,9 +48,14 @@ const StyledForm = styled.form`
 
 const Form = () => {
     const [state, setState] = useState<{
+        name?: string
+        consent?: boolean
+        email?: string
         feedback: string
         otherFeedback?: string
         messageSent?: boolean
+        errorMessage?: string
+
     }>({
         feedback: ''
     })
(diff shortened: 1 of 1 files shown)


gtsiolis (commented):
Hey @ghuntley! Here's some early feedback on this change. 

**suggestion:** Overall, I think it would be better to keep this form a) _anonymous_ and b) _optional_. Opening a new tab after installing or removing the extension is already a bit intrusive but I'm also in favor of keeping this flow. However, including form elements for user input like name and email could easily discourage users from considering filling out any inputs on the form.

If we still want to collect other user data like name and email, I'd vote for making it more clear that this data a) is _optional_ and b) only used in case a user is interested in hearing back from us for this specific submission. Even better, we could only show this form when a specific checkbox option is selected as it doesn't make sense for some options like _I prefer my local development_.

In terms of UX, I think we can improve a) the messaging, b) the tone of voice, and c) the checkbox options. Also, the text input elements introduced here could use better spacing (padding).

**thought**: A better alternative could be to proactively introduce help links where possible under each checkbox option or below the form. For example, linking directly to the screencasts, relevant documentation, etc. Many users still don't fully understand what dev environments are and this is a good opportunity for us to provide educational content or even suggest scheduling a call for help as included in the Free Trial email we're sending out.

Last but not least, we could use this PR only to fix #1009 and discuss this in a follow-up issue. 🏀

PR #1033, opened 2021-03-29
Fix plurals, change header

FeixuRuins:


FeixuRuins:
Another issue: "Claim Offer" buttons just redirect to `/subscription/`

